PR #1, opened 2022-11-10
init nixos-remote

Lassulus:
initial version of nixos-remote. takes a flake like https://github.com/Lassulus/flakes-testing to format/install

### DIFF
--- a/nixos-remote
+++ b/nixos-remote
@@ -87,6 +87,10 @@ fetch(){
     exit 1
   fi
 }
+if command -v tar >/dev/null 2>&1; then
+  echo "no tar command found, but required to unpack kexec tarball" >&2
+  exit 1
+fi
 rm -rf /root/kexec
 mkdir -p /root/kexec
 fetch "$kexec_url" | tar -C /root/kexec -xvzf-
--- a/nixos-remote
+++ b/nixos-remote
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+set -eufo pipefail
+set -x
+
+showUsage() {
+  cat <<USAGE
+Usage: $0 [options] ssh-host
+
+Options:
+
+* -f, --flake flake
+  set the flake to install the system from
+* --arg name value
+  pass value to nix-build. can be used to set disk-names for example
+* --argstr name value
+  pass value to nix-build as string
+USAGE
+}
+
+abort() {
+  echo "aborted: $*" >&2
+  exit 1
+}
+
+nix_args=()
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    -f | --flake)
+      flake=$2
+      shift
+      ;;
+    --argstr | --arg)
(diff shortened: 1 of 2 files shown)


nixos-remote
```diff
@@ -0,0 +1,116 @@
...
+    curl --fail -Ss -L "\$1"
+  elif command -v wget >/dev/null 2>&1; then
+    wget "\$1" -O-
+  else
+    echo "no downloader (curl or wget) found, bailing out"
+    exit 1
+  fi
+}
+rm -rf /root/kexec
+mkdir -p /root/kexec
```
Mic92, line 88: Maybe useful?

```suggestion
if command -v tar >/dev/null 2>&1; then
    echo "no tar command found, but required to unpack kexec tarball" >&2
    exit 1
fi
```

Lassulus:
possible next steps:
- non flake support (do we want this?)
- absorb https://github.com/numtide/terraform-nixos-install
- absorb https://github.com/nix-community/nixos-images
- documentation

Mic92:
I would say README is the most important one. Than I can have people starting to test this.

nixos-remote
Lassulus: alright, was absorbed

nixos-remote
```diff
@@ -0,0 +1,123 @@
...
+    wget "\$1" -O-
+  else
+    echo "no downloader (curl or wget) found, bailing out"
+    exit 1
+  fi
+}
+if command -v tar >/dev/null 2>&1; then
+  echo "no tar command found, but required to unpack kexec tarball" >&2
+  exit 1
+fi
```
Mic92, line 93: os=$(uname)
if [[ "$os" != "Linux" ]]; then
  echo "Expected Linux operating system, but got a '${os}'" >&2
  exit 1
fi

Lassulus: do we expect linux? does kexec not work on other OSes? :D

nixos-remote
```diff
@@ -0,0 +1,123 @@
...
+USAGE
+}
+
+abort() {
+  echo "aborted: $*" >&2
+  exit 1
+}
+
+nix_args=()
+kexec_url=https://github.com/nix-community/nixos-images/releases/download/nixos-22.05/nixos-kexec-installer-x86_64-linux.tar.gz
```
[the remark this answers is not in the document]
Lassulus: maybe we just make it dependant on uname -m from the target?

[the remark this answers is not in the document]
Mic92: target="$(uname -s).$(uname -m)"
case "$target" in
    Linux.x86_64)
        url="https://github.com/nix-community/nixos-images/releases/download/nixos-22.05/nixos-kexec-installer-x86_64-linux.tar.gz"
        ;;
     *)
       echo "$target is currently not supported by by our kexec tarball"
       exit 1
esac

PR #6, opened 2022-12-08
flakeify

Lassulus:


nixos-remote
```diff
@@ -94,14 +108,20 @@ if [[ ! -z "${flake+x}" ]]; then
   fi
   disko_script=$(nix_build "${flake}#nixosConfigurations.${flakeAttr}.config.system.build.disko")
   nixos_system=$(nix_build "${flake}#nixosConfigurations.${flakeAttr}.config.system.build.toplevel")
+elif [[ ! -z "${disko_script+x}" ]] && [[ ! -z "${nixos_system+x}" ]]; then
```
zimbatm, line 111:
```suggestion
elif [[ ! -z "${disko_script}" && ! -z "${nixos_system}" ]]; then
```
removing old POSIX sh relics

tests/from-nixos.nix
```diff
@@ -0,0 +1,129 @@
...
+        return machine
+
+    start_all()
+    installed.wait_for_unit("sshd.service")
+    installer.succeed("""
+      eval $(ssh-agent)
+      ssh-add /etc/sshKey
+      ${../nixos-remote} \
+        --no-ssh-copy-id \
+        --kexec ${kexec_tarball} \
```
Mic92, line 117: I was able to replace the import-from-derivation like this:

https://github.com/kuutamolabs/near-staking-knd/blob/f499d53a01ef3b3efdf69b0153aa975d37895769/nix/modules/tests/flake-module.nix#L8
https://github.com/kuutamolabs/near-staking-knd/blob/f499d53a01ef3b3efdf69b0153aa975d37895769/nix/modules/tests/install-nixos.nix#L116

PR #14, opened 2022-12-28
Add --disk-encryption-keys option

phaer:
Similar to `--extra-files`, but files are copied before partitioning and installation. This allows to specify i.e. disk encryption keys.

### DIFF
--- a/nixos-remote
+++ b/nixos-remote
@@ -214,7 +214,7 @@ if [[ -n ${disk_encryption_keys:-} ]]; then
   if [[ -d "$disk_encryption_keys" ]]; then
     disk_encryption_keys="$disk_encryption_keys/"
   fi
-  rsync -vaAXF -e "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" "$disk_encryption_keys" "${ssh_connection#ssh://}:/tmp/"
+  rsync -vrlF -e "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" "$disk_encryption_keys" "${ssh_connection#ssh://}:/tmp/"
 fi
 
 nixCopy --to "ssh://$ssh_connection" "$disko_script"
--- a/nixos-remote
+++ b/nixos-remote
@@ -22,6 +22,9 @@ Options:
   do not reboot after installation
 * --extra-files files
   files to copy into the new nixos installation
+* --disk-encryption-keys files
+  files to copy into the installer environment, after kexec but before installation. Can be 
+  used for things like disk encryption keys
 * --debug
   enable debug output
 USAGE
@@ -65,6 +68,10 @@ while [[ $# -gt 0 ]]; do
       ;;
     --extra-files)
       extra_files=$2
+      shift
+      ;;
+    --disk-encryption-keys)
+      disk_encryption_keys=$2
       shift
       ;;
     --stop-after-disko)
@@ -203,6 +210,13 @@ SSH
   until ssh_ -o ConnectTimeout=10 -- exit 0; do sleep 5; done
 fi
 
+if [[ -n ${disk_encryption_keys:-} ]]; then
+  if [[ -d "$disk_encryption_keys" ]]; then
+    disk_encryption_keys="$disk_encryption_keys/"
+  fi
+  rsync -vaAXF -e "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" "$disk_encryption_keys" "${ssh_connection#ssh://}:/tmp/"
+fi
+
 nixCopy --to "ssh://$ssh_connection" "$disko_script"
 ssh_ "$disko_script"
 
@@ -220,6 +234,8 @@ fi
 
 ssh_ <<SSH
 set -efu ${enable_debug}
+# needed for installation if initrd-secrets are used
+mkdir -p /mnt/tmp
 nixos-install --no-root-passwd --no-channel-copy --system "$nixos_system"
 ${maybereboot}
 SSH
--- a/nixos-remote
+++ b/nixos-remote
@@ -14,12 +14,12 @@ Options:
   if this is give, flake is not needed
 * --no-ssh-copy
   skip copying ssh-keys to target system
+* --no-reboot
+  do not reboot after installation, allowing further customization of the target installation.
 * --kexec url
   use another kexec tarball to bootstrap NixOS
 * --stop-after-disko
   exit after disko formating, you can then proceed to install manually or some other way
-* --no-reboot
-  do not reboot after installation
 * --extra-files files
   files to copy into the new nixos installation
 * --disk-encryption-keys files
--- a/nixos-remote
+++ b/nixos-remote
@@ -235,7 +235,7 @@ fi
 ssh_ <<SSH
 set -efu ${enable_debug}
 # needed for installation if initrd-secrets are used
-mkdir -p /mnt/tmp
+mkdir -m777 -p /mnt/tmp
 nixos-install --no-root-passwd --no-channel-copy --system "$nixos_system"
 ${maybereboot}
 SSH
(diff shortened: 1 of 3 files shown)


nixos-remote
```diff
@@ -222,6 +236,7 @@ set -efu
 if [ "$debug" = "y" ]; then
   set -x
 fi
+mkdir -p /mnt/tmp
```
Mic92, line 239:
```suggestion
# needed for installation if initrd-secrets are used
mkdir -m777 -p /mnt/tmp
```

nixos-remote
```diff
@@ -203,6 +210,13 @@ SSH
   until ssh_ -o ConnectTimeout=10 -- exit 0; do sleep 5; done
 fi
 
+if [[ -n ${disk_encryption_keys:-} ]]; then
```
Mic92, line 213: Is there some easy way add a test for this? It is probably not as commonly used and I want to avoid it to break like here: https://github.com/numtide/nixos-remote/pull/15#issuecomment-1367342213

Mic92: Some test that checks that this file exists in the installer is enough, you don't need to switch to disk encryption necessary.

phaer: As the file is only expected to exist in the kexec env after nixos-remote is run, I think we'd need a `--no-reboot` flag and do the reboot manually for that. Might be a good idea to allow extra customization (like chowning --extra-files) anyways.

Mic92: Yeah, I think `--no-reboot` is a good idea.

nixos-remote
```diff
@@ -222,6 +236,8 @@ set -efu
 if [ "$debug" = "y" ]; then
   set -x
 fi
+# needed for installation if initrd-secrets are used
+mkdir -p /mnt/tmp
```
Mic92, line 240: `/tmp` should be world-writeable. Otherwise random things will break (been there; done that)

```suggestion
mkdir -m777 -p /mnt/tmp
```

phaer:
I've added --reboot and tried to fix the test, but it's getting late and i'll need to leave it there for the holidays. 

I've tried to debug why machine.execute() seems to fail; I guess can't actually connect that simply to `installed`, after nixos-remote does its kexec via ssh.  I also tried various combinations of simply adding `ssh root@installed reboot`  to the block where nixos-remote is called, but it didn't work either. See log below

<details>
<summary>
nix run .#checks.x86_64-linux.from-nixos -L
</summary>
vm-test-run-nixos-remote> installer # + exit 0
vm-test-run-nixos-remote> (finished: must succeed:
vm-test-run-nixos-remote>   eval $(ssh-agent)
vm-test-run-nixos-remote>   ssh-add /etc/sshKey
vm-test-run-nixos-remote>   /nix/store/i80nyns4whpiy6jqrnzxpbbgl89cyw77-nixos-remote     --no-ssh-copy-id     --debug     --kexec /nix/store/srm04379s5p9rl6hnlq97bki8b2vy0j7-kexec-tarball/nixos-kexec-installer-x86_64-linux.tar.gz     --extra-files /tmp/extra-files     --disk-encryption-keys /tmp/disk-encryption-key     --store-paths /nix/store/9an6ka9jxv2275vm8hnbbz5xq39apjg9-disko /nix/store/miz679q04n6dwrk6hqmg92z67v3p5rmh-nixos-system-nixos-remote-23.05pre-git     --no-reboot     root@installed >&2
vm-test-run-nixos-remote> , in 95.36 seconds)
vm-test-run-nixos-remote> installed: waiting for the VM to finish booting
vm-test-run-nixos-remote> installed: connected to guest root shell
vm-test-run-nixos-remote> installed: (connecting took 0.00 seconds)
vm-test-run-nixos-remote> (finished: waiting for the VM to finish booting, in 0.00 seconds)
vm-test-run-nixos-remote> installed: must succeed: cat /tmp/disk-encryption-key
vm-test-run-nixos-remote> cleanup
vm-test-run-nixos-remote> kill machine (pid 8)
vm-test-run-nixos-remote> installed # [0;nixos@nixos: ~nixos@nixos:~]$ qemu-kvm: terminating on signal 15 from pid 6 (/nix/store/nzgqxw4d875pgl1lji8n3q82x3bf7ldi-python3-3.10.8/bin/python3.10)
vm-test-run-nixos-remote> kill machine (pid 19)
vm-test-run-nixos-remote> installer # qemu-kvm: terminating on signal 15 from pid 6 (/nix/store/nzgqxw4d875pgl1lji8n3q82x3bf7ldi-python3-3.10.8/bin/python3.10)
vm-test-run-nixos-remote> (finished: cleanup, in 0.09 seconds)
vm-test-run-nixos-remote> Traceback (most recent call last):
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/bin/.nixos-test-driver-wrapped", line 9, in <module>
vm-test-run-nixos-remote>     sys.exit(main())
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/lib/python3.10/site-packages/test_driver/__init__.py", line 114, in main
vm-test-run-nixos-remote>     driver.run_tests()
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/lib/python3.10/site-packages/test_driver/driver.py", line 146, in run_tests
vm-test-run-nixos-remote>     self.test_script()
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/lib/python3.10/site-packages/test_driver/driver.py", line 142, in test_script
vm-test-run-nixos-remote>     exec(self.tests, symbols, None)
vm-test-run-nixos-remote>   File "<string>", line 33, in <module>
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/lib/python3.10/site-packages/test_driver/machine.py", line 597, in succeed
vm-test-run-nixos-remote>     (status, out) = self.execute(command, timeout=timeout)
vm-test-run-nixos-remote>   File "/nix/store/fyhjj0p4h5lb9sqyll642h2yafv3rkzw-nixos-test-driver-1.1/lib/python3.10/site-packages/test_driver/machine.py", line 553, in execute
vm-test-run-nixos-remote>     rc = int(self._next_newline_closed_block_from_shell().strip())
vm-test-run-nixos-remote> ValueError: invalid literal for int() with base 10: 'echo ${PIPESTATUS[0]}'
vm-test-run-nixos-remote> kill vlan (pid 7)
error: builder for '/nix/store/8q2slmjqqihgb7pmi3z84315wl50zapd-vm-test-run-nixos-remote.drv' failed with exit code 1;
</details>

nixos-remote
phaer: Good catch, thanks! :100:

phaer:
Note to self: maybe we should use "cat | ssh" instead of rsync for disk encryption keys, so we can use bash file substition `<(command)` directly, without writing the key to a local file.

Mic92:
I rebased your pull request to fix the merge conflicts.

tests/from-nixos.nix
```diff
@@ -17,6 +17,7 @@
     start_all()
     installer.succeed("mkdir -p /tmp/extra-files/var/lib/secrets")
     installer.succeed("echo value > /tmp/extra-files/var/lib/secrets/key")
+    installer.succeed("echo super-secret > /tmp/disk-encryption-key")
```
Mic92, line 20: I also added a `tests/from-nixos-with-sudo.nix` test, which might be easier to extend as it does not reboot by default.

phaer: Great solution, thanks for finishing!

PR #15, opened 2022-12-29
Don't preserve local permissions for --extra-files

phaer:
This changes our rsync invocation so that the remote copies of --extra-files always belong to `root:root`. I believe that's a safer default than copying local permissions in all cases. Without this change, files which should belong to root would need to locally belong to root as well. In cases where one would like other permissions on the remote system, one could chown them after copying.

### DIFF
--- a/nixos-remote
+++ b/nixos-remote
@@ -211,7 +211,7 @@ if [[ -n ${extra_files:-} ]]; then
   if [[ -d "$extra_files" ]]; then
     extra_files="$extra_files/"
   fi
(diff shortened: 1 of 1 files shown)


Mic92:
I think some of flags where needed though. Now it is failing with: `after_install # cat: /var/lib/secrets/key: No such file or directory`

phaer:
Well that's a bit embarrassing tbh, I forgot that `-a` also implies `-rl`, only tested with a single file and didn't run the tests locally. Seems to work now

phaer:
Needed a rebase after the SSH connection changes on main, but conflicts are gone and tests are passing.

PR #20, opened 2023-01-04
allow pipes for --disk-encryption-keys...

phaer:
this allows passing multiple disk encryption keys, some of which might come in the form of unix pipes. 

It can be used with bash file substitution to pass a secret to the remote machine without writing it locally to disk.

Example

```
--disk-encryption-keys /tmp/disk-sda.key /tmp/just-a-file
--disk-encryption-keys /tmp/disk-sdb.key <(echo "top-secret") 
--disk-encryption-keys /root/another-path.key <(sops --extract '["disk-key"]' -d secrets/host.json) 
--disk-encryption-keys /tmp/disk-sdc <(sops -d secrets/disk-key.json) 
```

### DIFF
--- a/nixos-remote
+++ b/nixos-remote
@@ -216,7 +216,7 @@ fi
 for path in "${!disk_encryption_keys[@]}"
 do
   echo "Uploading ${disk_encryption_keys[$path]} to $path"
-  cat "${disk_encryption_keys[$path]}" | ssh_ "cat > $path"
+  ssh_ "cat > $path" < "${disk_encryption_keys[$path]}"
 done
 
 nixCopy --to "ssh://$ssh_connection" "$disko_script"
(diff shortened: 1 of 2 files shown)


nixos-remote
```diff
@@ -210,13 +213,11 @@ SSH
...
-if [[ -n ${disk_encryption_keys:-} ]]; then
-  if [[ -d "$disk_encryption_keys" ]]; then
-    disk_encryption_keys="$disk_encryption_keys/"
-  fi
-  rsync -vrlF -e "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" "$disk_encryption_keys" "${ssh_connection#ssh://}:/tmp/"
-fi
+for path in "${!disk_encryption_keys[@]}"
+do
+  echo "Uploading ${disk_encryption_keys[$path]} to $path"
+  ssh_ "cat > $path" < "${disk_encryption_keys[$path]}"
```
Mic92, line 219: Does anyone else beyond root needs to be able to read this?

```suggestion
  ssh_ "umask 077; cat > $path" < "${disk_encryption_keys[$path]}"
```

Mic92 (approved):
I like it. Only the one discussion point above.

PR #21, opened 2023-01-06
add terraform role and some minor changes

Lassulus:


terraform/main.tf
```diff
@@ -0,0 +1,8 @@
+resource "null_resource" "nixos-remote" {
+  provisioner "local-exec" {
+    environment = {
+      SSH_PRIVATE_KEY = var.ssh_private_key
+    }
+    command = "${path.module}/../nixos-remote --store-paths ${var.nixos_partitioner} ${var.nixos_system} ${var.target_user}@${var.target_host} --debug"
```
Mic92, line 6: Do we want `--debug` by default?
```suggestion
    command = "${path.module}/../nixos-remote --store-paths ${var.nixos_partitioner} ${var.nixos_system} ${var.target_user}@${var.target_host} --debug"
```

Lassulus: probably not, but we want it for debugging a WIP branch :)

src/nixos-remote.sh
```diff
@@ -155,11 +158,33 @@ else
   abort "flake must be set"
 fi
 
+if [[ -n ${SSH_PRIVATE_KEY:-} ]]; then
+  sshPrivateKeyFile=$(mktemp)
+  trap 'rm "$sshPrivateKeyFile"' EXIT
+  printf '%s' "$SSH_PRIVATE_KEY" >"$sshPrivateKeyFile"
+  chmod 0700 "$sshPrivateKeyFile"
```
Mic92, line 165:
```suggestion
```

src/nixos-remote.sh
```diff
@@ -98,12 +94,18 @@ while [[ $# -gt 0 ]]; do
   shift
 done
 
+# we generate a temporary ssh keypair that we can use during nixos-remote
+ssh_key_dir=$(mktemp -d)
+trap 'rm -rf "$ssh_key_dir"' EXIT
+mkdir -p "$ssh_key_dir"
+ssh-keygen -t ed25519 -f "$ssh_key_dir"/nixos-remote -P "" -C "nixos-remote" >/dev/null
```
phaer, line 101: Why do we require a temporary ssh key pair? I think it's debatable whether that's a "minor changes", but an explanatory comment would be appreciated :)

src/nixos-remote.sh
```diff
@@ -155,11 +158,33 @@ else
   abort "flake must be set"
 fi
 
+if [[ -n ${SSH_PRIVATE_KEY:-} ]]; then
+  sshPrivateKeyFile=$(mktemp)
+  trap 'rm "$sshPrivateKeyFile"' EXIT
+  printf '%s' "$SSH_PRIVATE_KEY" >"$sshPrivateKeyFile"
+  chmod 0700 "$sshPrivateKeyFile"
+  unset SSH_AUTH_SOCK # don't use system agent if key was supplied
```
phaer, line 166: This comment seems obsolete if we always use the temporary key.

terraform/install/main.tf
```diff
@@ -0,0 +1,8 @@
+resource "null_resource" "nixos-remote" {
+  provisioner "local-exec" {
+    environment = {
+      SSH_PRIVATE_KEY = var.ssh_private_key
+    }
+    command = "${path.module}/../../src/nixos-remote.sh --store-paths ${var.nixos_partitioner} ${var.nixos_system} ${var.target_user}@${var.target_host}"
```
phaer, line 6: `nix run ${path.module}#nixos-remote --` should be preferred as it ensures the right dependencies

src/nixos-remote.sh
Lassulus: we still use SSH_AUTH_SOCK for ssh-copy-id if no ssh-key was specified, so this comment should still be correct

src/nixos-remote.sh
Lassulus: to have less branching in the code, if we want to supply optional keys, we would need to wrap every call of ssh, ssh-copy-id etc with the supplied private key, also we have to keep track which public keys to copy to the target system and ssh-copy-id can't copy public keys of private keys which are passed via environment.

Mic92: It also makes it easier to support a usecase where you people have not set up a public ssh key but need to type in the password. This way they only need to do this once.

phaer: > we would need to wrap every call of ssh, ssh-copy-id etc with the supplied private key

I can take this on in a separate PR, it shouldn't block anything here. But I think something like $SSH_ARGS everywhere is preferable to a mandatory auto-generated key.

> also we have to keep track which public keys to copy to the target system and ssh-copy-id can't copy public keys of private keys which are passed via environment.

ssh-keygen can derive a public key from a given private key, so that should work?

> It also makes it easier to support a usecase where you people have not set up a public ssh key but need to type in the password. This way they only need to do this once

True, that's a nice feature to have - but i still think auto-generated credentials should be either optional via a flag or at least explicit (mentioned in README or log output).

src/nixos-remote.sh
phaer: Thanks, missed that

src/nixos-remote.sh
Lassulus: well the key won't survive the install anyway, we could mention that more I guess. SSH_ARGS was also something I did first, but the edgecases are really building up for different usecases (supply a private key via env, supply a private key via agent, use password login)

Mic92: I got to say I am actually with @Lassulus here.
With nixos-remote we also want to build a solution for newcomers and less-experienced users.
Less-experienced users often face issues with key handling and this seems one solution that should work in every use case while not adding a lot of code. 

@phaer is it just the overhead of generating a key here that is your concerned about? I think ed25519 keys should only take a small portion of the time it would take to evaluate any nixos configuration:

```
$ time ssh-keygen -t ed25519 -f dev1-key -N ''
Generating public/private ed25519 key pair.
Your identification has been saved in dev1-key
Your public key has been saved in dev1-key.pub
The key fingerprint is:
SHA256:6V35HMMSkegPXwz9ngzAzZNLzh7FmtzHMMIeaZUrLJ4 joerg@turingmachine
The key's randomart image is:
+--[ED25519 256]--+
|          .o+=+. |
|          .oOB=o |
|         . +B=B* |
|         .+ +&+.+|
|        S. *=oB.o|
|       . .E.o= =.|
|        . .   o  |
|                 |
|                 |
+----[SHA256]-----+
ssh-keygen -t ed25519 -f dev1-key -N ''  0,00s user 0,01s system 94% cpu 0,010 total
```

^ I would argue that this computational overhead is smaller than the extra shell code that would need to be added to cover all use cases that we might run into.

phaer: As discussed in chat: this is fine as it is, we can make it optional or more explicit after the terraform integration is up and running.

Lassulus:
bors merge

Kranzes:
I wonder if this is something we can write a test for.

terraform/nix-build/nix-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -efu
+
+declare file attribute
+eval "$(jq -r '@sh "attribute=\(.attribute) file=\(.file)"')"
```
Kranzes, line 5: Where is jq and other dependencies declared?

Lassulus: I guess in the environment terraform is run from. not sure how we could take easily care of that. maybe a nix-shell shebang. but that would also increase clousure size

Kranzes: Increasing closure size for the sake of purity is totally fine.

phaer: I am in favor of keeping dependencies explicit, but this PR as it has been merged also assumes that `terraform` is already in path and does not lock the external provider (hashicorp/null) at all, unless i am mistaken?

Maybe it would be a good compromise to offer a devShell with terraform, the provider, jq and what else might be needed?

terraform/install/main.tf
```diff
@@ -0,0 +1,8 @@
+resource "null_resource" "nixos-remote" {
+  provisioner "local-exec" {
+    environment = {
+      SSH_PRIVATE_KEY = var.ssh_private_key
+    }
+    command = "nix run ${path.module}#nixos-remote -- --store-paths ${var.nixos_partitioner} ${var.nixos_system} ${var.target_user}@${var.target_host}"
```
Mic92, line 6: Should be nixos-anywhere?

terraform/nixos-rebuild/deploy.sh
```diff
@@ -0,0 +1,34 @@
...
+  echo "$SSH_KEY" >"$sshPrivateKeyFile"
+  chmod 0700 "$sshPrivateKeyFile"
+  unset SSH_AUTH_SOCK # don't use system agent if key was supplied
+  sshOpts+=(-o "IdentityFile=${sshPrivateKeyFile}")
+fi
+
+NIX_SSHOPTS="${sshOpts[*]}" retry -t 10 -d 10 -- nix copy -s --experimental-features nix-command --to "ssh://$TARGET_HOST" "$NIXOS_SYSTEM"
+
+# shellcheck disable=SC2029
+ssh "${sshOpts[@]}" "$TARGET_HOST" "nix-env -p /nix/var/nix/profiles/system --set $(printf "%q" "$NIXOS_SYSTEM"); /nix/var/nix/profiles/system/bin/switch-to-configuration switch" || :
```
Mic92, line 34: This is not a proper replacement for `nixos-rebuild` i.e. it does not use the nix version specified in the nixos configuration. I would rather suggest to use `nixos-rebuild --target-host <ip> --build-host` instead of re-implementing it in suboptimal way.

Mic92: than's evaled twice I guess... Anyway I would personally use this for my projects.

Lassulus: yeah, I mainly used nix-copy-closure directly so we can support ssh:// style links. since they allow easy deployment on alternate ports. maybe we can add that functionality to nixos-rebuild and change our code then?

PR #22, opened 2023-01-10
Use writeShellApplication for packaging

Kranzes:
With this change the packaging will be made with `writeShellApplication` which allows adding runtime inputs (deps) to the script by setting the $PATH variable at the top of the file. It also has a check phase which tests the shell file against `shellcheck` like we did before manually. It also sets the safety bash options.

you can read more about `writeShellApplication` here -> https://github.com/NixOS/nixpkgs/blob/d3a307c813cbd209fe1e13ac9246b3fa3ea93c12/pkgs/build-support/trivial-builders.nix#L282


src/nixos-remote.sh
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env bash
-set -eufo pipefail
-
```
zimbatm, line 3: This header is useful for shellcheck and editor completion so they know that it's a bash script. Is it necessary to remove it?

Kranzes: I added an file extension as that's another way for an editor to detect the file type. I removed them because writeShellApplication adds them already.

Mic92: Why was the `set -eufo pipefail` removed?

Kranzes: because writeShellApplication adds it for you, and you would rarely run the shell file directly unpackaged because that would be impure since it won't include the dependencies listed in the derivation.

Kranzes:
How should I go about bringing the flake's package output into scope in `tests/from-nixos.nix` for example?

Kranzes:
I have a fork/branch for with some more improvements (& that utilizes of [flake.parts](http://flake.parts/)), is that something you are interested in me PR'ing over?

src/nixos-remote.sh
Mic92: Ah, I was looking at the wrong function.

Mic92:
bors merge

Mic92:
> I have a fork/branch for with some more improvements (& that utilizes of [flake.parts](http://flake.parts/)), is that something you are interested in me PR'ing over?

Yes. Looks useful: https://github.com/Kranzes/nixos-remote/commit/b400913411bf42a49479abd50c583935f704a0cb

Mic92:
But maybe keep the packages with importable with `callPackage`. I think some folks still want to manually import them without using flakes.

Kranzes:
> But maybe keep the packages with importable with `callPackage`. I think some folks still want to manually import them without using flakes.

what about using `builtins.getFlake`?

Kranzes:
> > I have a fork/branch for with some more improvements (& that utilizes of [flake.parts](http://flake.parts/)), is that something you are interested in me PR'ing over?
> 
> Yes. Looks useful: [Kranzes@b400913](https://github.com/Kranzes/nixos-remote/commit/b400913411bf42a49479abd50c583935f704a0cb)

I created the PR: #23

Mic92:
nixos-remote can be still used without flakes. All parameter have fallbacks for store paths.

PR #23, opened 2023-01-10
nix: refactor and start using flake-parts

Kranzes:


src/default.nix
```diff
@@ -1,20 +1,19 @@
-{ writeShellApplication
-, openssh
-, gitMinimal
-, rsync
-, nix
-, coreutils
-, curl
-}:
```
Mic92, line 8: I would keep this a package and only have flake-parts in flake.nix for now. This allows to still override all inputs using `.override` and people still can import this repository with niv in case they don't want to use flakes.

tests/default.nix
```diff
@@ -0,0 +1,17 @@
+{ withSystem, inputs, ... }:
+
+{
+  flake.checks.x86_64-linux = withSystem "x86_64-linux" ({ pkgs, system, inputs', config, ... }:
```
Mic92, line 4: I guess here its fine to have a flake-module... I would actually call it `flake-module.nix` to make it more clear what it is.

Mic92, line 4: Now that we have flake-parts we can also at least add aarch64-linux support for checks as well.

Kranzes: Does nixos-remote works on darwin? if not we can just add this to perSystem.

Mic92: If you have a linux remote builder yes, but not for tests.

Kranzes: When I use flake-parts for me the `default.nix` is usually the root where I define the flake module/outputs that then import the other nix files in the sub directory. It's a matter of naming really, up to you, tell me what you prefer.

Kranzes: The reason I am asking is to know whether the $system for packages can match the one for checks.

Mic92: I prefer having flake-module's recognizable from the name to make them easier to distinguish from other files that might live in the same directory.

Kranzes: In other words, can you do `nix build .#packages.aarch64-darwin.nixos-remote` on a mac?

src/default.nix
Kranzes: Do I do the same for the docs?

tests/default.nix
Mic92: You can do `nix build .#packages.aarch64-darwin.nixos-remote` on a mac and the script should work

src/default.nix
Mic92: I think there it might be not needed, so no strong opinion.

Kranzes: I don't even think it can be overwritten anyway.

tests/default.nix
Kranzes: for docs/default.nix do I also change it to docs/flake-module.nix?

Kranzes: Looks like nixos-images only builds for `x86_64-linux` anyway so I guess I will keep it the way it is for now.

Kranzes:
@Mic92 I pushed with what you suggested, tell me if there is anything else I missed.

tests/default.nix
Kranzes: Also Garnix's aarch64-linux support isn't mediocre at best, it uses emulation. I personally use hercules-ci which lets me use my own runners/agents.

phaer (approved):
Thanks, LGTM! Builds (with flakes on x86_64-linux) and I think flake-parts are an improvement. No strong opinion regarding the aarch64/ci discussion.

tests/default.nix
Mic92: I actually already used it for aarch64 by building the image myself locally (see project readme). However I might also switch to hercules ci that we already deployed in nix-community.

Mic92:
bors merge

PR #24, opened 2023-01-10
README: suggest the use of the packaged script

Kranzes:
Without this change users will be running the script directly and might encounter errors as they are using the script unpackaged without its dependencies.


zimbatm:
bors merge

zimbatm (approved):

PR #26, opened 2023-01-11
nixos-remote.sh: don't create symlink in dir & add --print-build-logs/-L flag

Kranzes:
- For the sake of consistency I've renamed the function `nixCopy` to `nix_copy`.
- Again, for for consistency I made `nix copy` and `nix build` have the same experimental features enabled.
- For `nix_build` it will no longer keep/create a `result` symlink in the current directory. Essentially our use for `nix build` is like the old CLI's `nix-store --realise`.
- Added the flag -L/--print-build-logs which passes this same flag to all invocations of `nix` via the `nix_options` bash array. This flag is also enabled by default when using the `--debug` flag.

### DIFF
--- a/src/nixos-remote.sh
+++ b/src/nixos-remote.sh
@@ -6,6 +6,8 @@ Options:
 
 * -f, --flake flake
   set the flake to install the system from
+* -L, --print-build-logs
+  print full build logs
 * -s, --store-paths
   set the store paths to the disko-script and nixos-system directly
   if this is give, flake is not needed
@@ -44,6 +46,9 @@ while [[ $# -gt 0 ]]; do
       flake=$2
       shift
       ;;
+    -L | --print-build-logs)
+      print_build_logs=y
+      ;;
     -s | --store-paths)
       disko_script=$(readlink -f "$2")
       nixos_system=$(readlink -f "$3")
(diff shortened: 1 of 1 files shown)


src/nixos-remote.sh
```diff
@@ -100,15 +100,17 @@ timeout_ssh_() {
...
+nix_copy() {
+  NIX_SSHOPTS='-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no' nix copy \
+    --extra-experimental-features 'nix-command flakes' \
+    "$@"
 }
 nix_build() {
-  nix \
-    --experimental-features flakes build \
-    --extra-experimental-features nix-command \
+  nix build \
```
Mic92, line 109: Potentially we want to more logs here in case something goes wrong with the build:

```suggestion
  nix build -L \
```

Mic92:
I think @Lassulus also was about to do something about making nixos-remote using the binary cache...

src/nixos-remote.sh
Kranzes: I did notice nixos-remote output is either really loud and useless or super silent and I don't actually know if anything is happening. Maybe printing nix build logs should be a flag?

Mic92: We can also leave this for another PR...

Kranzes: sure, just merge this how it is and I will make a follow-up PR soon.

Mic92:
bors merge

Mic92:
bors merge

Mic92:
bors merge

Mic92:
bors merge

PR #32, opened 2023-01-12
treewide: add formatting via treefmt

Kranzes:
This seems like a good idea to have, no strong opinions though.


Mic92:
bors merge

PR #33, opened 2023-01-12
perform reboot in the background

Mic92:
This allows the user to monitor reboot happens from a script and run further actions after it comes back.


Lassulus (approved):

Mic92:
bors merge

PR #35, opened 2023-01-15
extra-files: use rsync -p, -FF

phaer:
After removing -a in #15, remote extra files are implictly chowned according to the remote umask. So we (re-)add --perms to apply local file permissions remotely as this should mostly be expected behavior.

The second -F disables uploading the .rsync-filter file itself if it exists.


phaer:
ci-failure looks unrelated to this change?

Lassulus (approved):

phaer:
bors merge

PR #36, opened 2023-01-15
run nix copy --substitute-on-destination by default

phaer:
I think that's a good default as downloading from cache.nixos.org is often at least as fast as uploading from the machine where nixos-remote runs and often much faster if running nixos-remote over a typical residential internet connection.

On my home network connection this reduces the whole time it takes to deploy a new hcloud machine from my workstation from ~11min to ~3min.


Mic92:
bors merge

PR #37, opened 2023-01-15
fix maybesudo quotation

phaer:
this specific form did not work for me on machines without sudo:

```
+ '' rm -rf /root/kexec
-bash: line 2: : command not found
```


Mic92:
bors merge

PR #39, opened 2023-01-17
detect nixos installer

Mic92:
This requires nixos unstable

### DIFF
--- a/src/nixos-remote.sh
+++ b/src/nixos-remote.sh
@@ -172,9 +172,12 @@ set -efu ${enable_debug}
 has(){
   command -v "\$1" >/dev/null && echo "y" || echo "n"
 }
+is_nixos=\$(if test -f /etc/os-release && grep -q ID=nixos /etc/os-release; then echo "y"; else echo "n"; fi)
 cat <<FACTS
 is_os=\$(uname)
 is_kexec=\$(if test -f /etc/is_kexec; then echo "y"; else echo "n"; fi)
+is_nixos=\$is_nixos
+is_installer=\$(if [[ "\$is_nixos" == "y" ]] && grep -q VARIANT_ID=installer /etc/os-release; then echo "y"; else echo "n"; fi)
 has_tar=\$(has tar)
 has_sudo=\$(has sudo)
 has_wget=\$(has wget)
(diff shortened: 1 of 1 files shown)


src/nixos-remote.sh
```diff
@@ -204,7 +207,7 @@ if [[ ${is_kexec-n} != "y" ]] && [[ ${no_ssh_copy-n} != "y" ]]; then
   ssh-copy-id -o ConnectTimeout=10 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no "$ssh_connection"
 fi
 
-if [[ ${is_kexec-n} == "n" ]]; then
+if [[ ${is_kexec-n} == "n" ]] && [[ ${is_installer-n} == "n" ]]; then
```
phaer, line 210: Shouldn't it be sufficient to just check for is_installer and remove is_kexec here?

Mic92: Since we already using this in production, it is just nice to have for the transition period.

phaer (approved):

src/nixos-remote.sh
```diff
@@ -172,9 +172,12 @@ set -efu ${enable_debug}
 has(){
   command -v "\$1" >/dev/null && echo "y" || echo "n"
 }
+is_nixos=\$(if test -f /etc/os-release && grep -q ID=nixos /etc/os-release; then echo "y"; else echo "n"; fi)
 cat <<FACTS
 is_os=\$(uname)
 is_kexec=\$(if test -f /etc/is_kexec; then echo "y"; else echo "n"; fi)
+is_nixos=\$is_nixos
+is_installer=\$(if [[ "\$is_nixos" == "y" ]] && grep -q VARIANT_ID=installer /etc/os-release; then echo "y"; else echo "n"; fi)
```
dasJ, line 180: How was this tested? We quote the values in /etc/os-release since https://github.com/NixOS/nixpkgs/commit/bae181d3f0f453d9a23cf5e899c2cb0f96e91fef, so this will never match

dasJ: I see, it's because of `escapeIfNecessary`, sorry for the noise.

PR #42, opened 2023-01-18
accept flake config by default

Mic92:
It seems unlikely that we deploy untrusted flakes, so this seems like a sensible default.

### DIFF
--- a/src/nixos-remote.sh
+++ b/src/nixos-remote.sh
@@ -133,6 +133,7 @@ nix_build() {
     "${nix_options[@]}" \
     --print-out-paths \
     --no-link \
+    --accept-flake-config \
     "$@"
 }
 


phaer:
So my nix 2.11 doesn't have this flag yet, but as i understand it, this trusts settings such as 

``
    nixConfig.extra-substituters = [ "https://colmena.cachix.org" ];
    nixConfig.extra-trusted-public-keys = [ "colmena.cachix.org-1:7BzpDnjjH8ki2CT3f6GdOk7QAzPOl+1t3LvTLXqYcSg=" ];
``

on your flake, without prompting the user? I believe the default is to prompt only the first time and then remember it in `~/.local/share/nix/trusted-settings.json`.

If so, i'd prefer to not have it in here, as it might implicitly add trusted substituters to the system running nixos-remote, not the system to install; a delicate matter on workstations. 

An alternative could be to just add "--nix-options" to allow users passing it themselves? (Or just an opt-in-flag --accept-flake-config)

phaer:
Is --option from #44 sufficient as a replacement for this, or do you want to merge this after discussion in chat? :)

PR #46, opened 2023-01-22
handle error case better when facts return no results back

Mic92:


src/nixos-remote.sh
```diff
@@ -50,6 +50,8 @@ nix_options=(
 
 declare -A disk_encryption_keys
 
+set -eu -o pipefail
```
Kranzes, line 53: This is added to the wrapped script with writeShellApplication, it probably doesn't harm to have it twice but what is the reason for doing it?

Mic92: I forgot that this is already done there.

phaer (commented):
Aborting when fact collection fails is surely a good idea.

But whats up with removing the colons? If i am not mistaken, the only difference the colon would make would be to handle i.e. empty strings the same as undefined variable. Isn't that the right thing to do, at least for flakeAttrs & so?

Mic92:
> Aborting when fact collection fails is surely a good idea.
> 
> But whats up with removing the colons? If i am not mistaken, the only difference the colon would make would be to handle i.e. empty strings the same as undefined variable. Isn't that the right thing to do, at least for flakeAttrs & so?

shfmt did this when adding `set -eux -o pipefail`, so I assume it's fine.

phaer:
Oh okay, thought that was a deliberate choice. But shouldn't matter too much in this case, because one would hopefully not set them to a null value

phaer:
bors merge

Mic92:
bors merge

Mic92:
bors merge

Mic92:
bors merge

Mic92:
bors merge

PR #47, opened 2023-01-22
add --from option

Mic92:
This is useful if your binary cache already has pre-build the machine.

### DIFF
--- a/src/nixos-remote.sh
+++ b/src/nixos-remote.sh
@@ -30,6 +30,8 @@ Options:
   enable debug output
 * --option KEY VALUE
   nix option to pass to every nix related command
+* --from store-uri
+  URL of the source Nix store to copy the nixos and disko closure from
 USAGE
 }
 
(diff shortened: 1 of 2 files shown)


src/nixos-remote.sh
```diff
@@ -104,7 +110,7 @@ while [[ $# -gt 0 ]]; do
     nix_options+=("--option" "$key" "$value")
     ;;
   --no-substitute-on-destination)
-    substitute_on_destination=""
+    nix_copy_options+=("--substitute-on-destination")
```
phaer, line 113: Either I am missing something, or this does the opposite of what the option implies. It *adds* --substitute-on-destination if i pass --no... to nixos-remote right?

Mic92: yes, you are right.

Mic92: Reworked.

phaer (approved):

Mic92:
bors merge

phaer:
Haha, that was quick - was just about to ask you whether its good to merge, as the last push was just 1 min ago :D

PR #50, opened 2023-01-31
wip: pxe boot

Mic92:
Depends on https://github.com/nix-community/nixos-images/pull/66 (I might actually move this images to the nixos-remote repo instead)

### DIFF
--- a/nixos-remote-pxe.py
+++ b/nixos-remote-pxe.py
@@ -1,0 +1,419 @@
+#!/usr/bin/env python3
+
+import subprocess
+import argparse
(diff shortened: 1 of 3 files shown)


pogobanane:
when restarting the script (just to get dhcp again or because some nix build failed) the ssh key changes which is annoying.

Mic92:
I think the build should actually happen before it starts and dnsmasq/pixicore. This makes restarts less often needed in this case. Can you elaborate why you needed to "get DHCP again"?

PR #71, opened 2023-02-14
terraform: add dependencies to script via shebang

Kranzes:


blaggacao:
Seems related to https://github.com/numtide/nixos-anywhere/issues/2. Which kight even already be implemented?

phaer (approved):
LGTM since inputs seem to implicitly taken from the global env without this?

terraform/nixos-rebuild/deploy.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env -S nix shell --inputs-from ../../ nixpkgs#openssh nixpkgs#nix nixpkgs#coreutils nixpkgs#runtimeShellPackage -c bash
```
Mic92, line 1: I think this should also enable experiment flags for nix-command and flakes in case the user does not have this enabled.

also:

```suggestion
#!/usr/bin/env -S nix shell --inputs-from path:./../ nixpkgs#openssh nixpkgs#nix nixpkgs#coreutils nixpkgs#runtimeShellPackage -c bash
```

otherwise it might discover a git repository and the wrong flake.

Mic92: we had issues vendoring openssh in the past: https://github.com/numtide/nixos-anywhere/issues/62
I think openssh should be only provided as a fallback.

Mic92: one idea: `export PATH=$PATH:$(nix shell --pure ... -c sh -c 'echo $PATH')`

terraform/nixos-rebuild/deploy.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env -S nix --extra-experimental-features 'nix-command flakes' shell --inputs-from path:../../ nixpkgs#nix nixpkgs#coreutils nixpkgs#runtimeShellPackage -c bash
```
Mic92, line 1: Does this work on macOS?

Mic92: Looks like it should: https://ss64.com/osx/env.html

jfroche, line 1: This change fails for me as `path:../../` assume that I have a flake.nix in my project parent directory.
I am have tested calling your change using:
```terraform
module "deploy_vault" {
  source                 = "github.com/Kranzes/nixos-remote//terraform/all-in-one"
  ...
}
```

The path in `--inputs-from` should be relative to the terraform variable `path.module` to get the flake.nix in nixos-anywhere root

phaer: @Kranzes Are you still interested in this? I think with the change suggested above, we should be good to merge :)

Kranzes: I don't use this terraform module and can't actually test this. Feel free to continue on your own.

PR #72, opened 2023-02-22
Terraform module fixes

Mic92:


terraform/nixos-rebuild/deploy.sh
```diff
@@ -28,7 +28,13 @@ if [[ -n ${SSH_KEY+x} && ${SSH_KEY} != "-" ]]; then
   sshOpts+=(-o "IdentityFile=${sshPrivateKeyFile}")
 fi
 
-NIX_SSHOPTS="${sshOpts[*]}" retry -t 10 -d 10 -- nix copy -s --experimental-features nix-command --to "ssh://$TARGET_HOST" "$NIXOS_SYSTEM"
```
Lassulus, line 31: not sure why you don't like retry. but I would reimplement it with until if you want to have less dependencies?

Mic92: It was just a bit of a sneaky dependency that you don't see until nixos-rebuild runs. I just wanted to avoid having to add this package to every project

phaer (commented):
LGTM from a quick read, but I haven't tested the terraform integration yet myself

terraform/nixos-rebuild/deploy.sh
Mic92: What would the `until` implementation looks like?

Lassulus:
```
try=1
until NIX_SSHOPTS="${sshOpts[*]}" nix copy -s --experimental-features nix-command --to "ssh://$TARGET_HOST" "$NIXOS_SYSTEM"; do
  if [[ $try > 10 ]]; then
    exit 1
  fi
  sleep 10
  try=$((try + 1))
done
```

your code also silently fails, but maybe that's not an issue

Mic92: done.

Mic92:
Tested sufficiently in production.

Mic92:
bors merge

PR #77, opened 2023-03-04
wait for machine to be shutdown after deploy

Mic92:
This is useful for automation i.e. our terraform module tries to run nixos-rebuild against the nixos installer instead of the actual machine.


Lassulus (approved):

Lassulus:
bors merge

PR #78, opened 2023-03-21
terraform: allow nixos-rebuild to use specified private key for deployment

jfroche:
`nixos-rebuild/deploy.sh` script enable ssh authentication with a given private key through the `SSH_KEY` environment variable.

This variable is set to the private key used for the initial deployment.


terraform/all-in-one/main.tf
```diff
@@ -33,6 +33,7 @@ module "nixos-rebuild" {
   ]
   source = "../nixos-rebuild"
   nixos_system = module.system-build.result.out
+  ssh_private_key = var.ssh_private_key
```
Mic92, line 36: I think we should have two variables here: `install_ssh_key` and `deployment_ssh_key` so that they can be set independently depending on the use case.
It's often not good practice to have ssh keys as string in terraform as it leaks the key into the state and having the differentiation allows users to limit the key to just installation time.

terraform/nixos-rebuild/variables.tf
```diff
@@ -19,3 +19,9 @@ variable "target_port" {
   description = "SSH port used to connect to the target_host"
   default     = 22
 }
+
+variable "ssh_private_key" {
+  type        = string
+  description = "Content of private key used to connect to the target_host"
+  default     = ""
```
Mic92, line 26: Otherwise the variable is `set` according to deploy.sh: `test -n ${SSH_KEY+x} && echo yes`

```suggestion
  default     = "-"
```

terraform/nixos-rebuild/variables.tf
```diff
@@ -19,3 +19,9 @@ variable "target_port" {
   description = "SSH port used to connect to the target_host"
   default     = 22
 }
+
+variable "ssh_private_key" {
+  type        = string
+  description = "Content of private key used to connect to the target_host"
```
Mic92, line 25:
```suggestion
  description = "Content of private key used to connect to the target_host. If set to - no key is passed to openssh and ssh will back to its own configuration".
```

terraform/nixos-rebuild/variables.tf
jfroche: Good catch, thank you !

srounce:
Have just run this with `install_user`/`install_ssh_key` set and `target_user`/`deployment_ssh_key` left unset. Can confirm this case works as expected.

Lassulus (approved):

terraform/nixos-rebuild/variables.tf
```diff
@@ -19,3 +19,9 @@ variable "target_port" {
   description = "SSH port used to connect to the target_host"
   default     = 22
 }
+
+variable "ssh_private_key" {
+  type        = string
+  description = "Content of private key used to connect to the target_host. If set to - no key is passed to openssh and ssh will back to its own configuration"
+  default     = "-"
```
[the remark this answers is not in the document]
Lassulus: because not setting the value is more complicated in terraform then setting a dummy value which basically means: "no value"

Lassulus:
bors merge

PR #80, opened 2023-03-31
fix(rebuild): use sudo if rebuild with another user than root

jfroche:
In order to switch the NixOS configuration, we must have root access. If we are connecting to the target host with a non-root user, we use the "sudo" command.

We could instead use `nixos-rebuild --use-remote-sudo --target-host` but it would evaluate the nixos system again.


Mic92 (approved):
Can be merged when tested.

jfroche:
bors merge

PR #86, opened 2023-04-25
add --build-on-remote flag

Lassulus:


src/nixos-anywhere.sh
```diff
@@ -173,8 +178,10 @@ if [[ -n ${flake-} ]]; then
     echo 'For example, to use the output nixosConfigurations.foo from the flake.nix, append "#foo" to the flake-uri.' >&2
     exit 1
   fi
-  disko_script=$(nix_build "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.disko")
-  nixos_system=$(nix_build "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.toplevel")
+  if [[ ${build_on_remote-n} == "n" ]]; then
+    disko_script=$(nix_build "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.disko")
+    nixos_system=$(nix_build "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.toplevel")
```
Mic92, line 183:
```suggestion
    nixos_system=$(nix_build "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.toplevel")
    elif [[ -z "${flake}" ]]; then
      abort "--build-on-remote also requires --flake to be set"
```

Lassulus: well it doesn't need flakes. but it has no effect without flakes, since the paths will already be build

Mic92: if there are no flakes how does it work? I don't see where machines are build on the remote?

Mic92: Ah. I right I thought we also supported non-flake options but we actually do not for nixos-anywhere.

Mic92:
bors merge

Mic92 (approved):

PR #98, opened 2023-05-10
src/nixos-anywhere.sh: add shebang

zimbatm:
It's not strictly needed because Nix wraps it with writeShellApplication, but it makes it easier for the script to be standalone and use the right syntax highlighter in shells.


Mic92:
bors merge

Mic92:
the nix sandbox has /bin/sh but not /usr/bin/env afaik.

PR #99, opened 2023-05-10
add -i to allow passing private key files

zimbatm:
Mirror the `-i` option from SSH, so you can run `nixos-anywhere ~/.ssh/other_key`.

This commit also fixes an issue where the generated key-pair would stay
around when using the SSH_PRIVATE_KEY env var.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -7,8 +7,10 @@ Usage: nixos-anywhere [options] ssh-host
 
 Options:
 
-* -f, --flake flake
-  set the flake to install the system from
+* -f, --flake <flake_uri>
+  set the flake to install the system from.
+* -i <identity_file>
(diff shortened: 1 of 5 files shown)


phaer (approved):

Lassulus:
ah but the test seems unhappy

Mic92:
bors merge

src/nixos-anywhere.sh
```diff
@@ -67,6 +70,10 @@ while [[ $# -gt 0 ]]; do
     flake=$2
     shift
     ;;
+  -i)
+    ssh_private_key_file=$2
```
Mic92, line 74: I changed this from $1 because it was receiving the wrong argument.

Mic92:
bors merge

Mic92:
bors merge

PR #111, opened 2023-05-15
fix rsync changing permissions

Mic92:


Mic92:
Extracted from https://github.com/numtide/nixos-anywhere/pull/50 but I also had this issue in other instances.

PR #113, opened 2023-05-16
Added Quickstart guide

JillThornhill:
Populated the Quickstart guide


Mic92:
Example looks good.
I will simplify the example in the README a bit. Especially so that people not use it without adapting it to their system.

bors merge

PR #115, opened 2023-05-25
More changes to documentation

JillThornhill:
1. Tidying up, adding index and images
2. Edits to the Howto, including changing examples to run from Numtide repo
3. Added error messages to the Reference
4. Created the beginnings of a Support Matrix; needs much more info


phaer (approved):
LGTM!

docs/supportmatrix.md
```diff
@@ -0,0 +1,29 @@
...
+
+We have verified that **nixos-anywhere** runs successfully on the environments listed below. 
+
+**kexec boot:**    Indicates whether it can work by using a kexec boot
+
+**Image boot:**   Indicates whether it can work if booted from a NixOS installer
+
+**Own kexec:**     Indicates whether it can work if you supply your own kexec 
+
+| **Platform**  | **Operating System** | kexec boot | Image boot | Own kexec |
```
dantefromhell, line 19: I have done a bunch of successful testing in the OVH Cloud and I'm happy to add my findings.

However I have experienced differences in results depending on machine type (e.g. 2GB RAM failing until commit 848a9cd came along), hence I am wondering if the table should be a bit more verbose about machine specs.

Another factor might be arying CPU architectures - e.g. Hetzner Cloud now offers Intel and AMD CPUs.

Maybe the solution could be as easy as adding a row for comments that either lays out successfully tested machine types or those known to fail?

JillThornhill: Maybe I can open a doc somewhere like HackMD where we can edit, comment, play with the format and find something that works; also people can add their findings to it? I'll create something tomorrow and post a link

JillThornhill: I've put this onto HackMD here: https://hackmd.io/2i_NRaiRQluGm_YXHAAaTQ?both

Maybe we can use this to play around, decide on the format we need and add more platforms etc. Once done I can bring it back into Github and create a pull request

zimbatm: This can be done in a subsequent PR

zimbatm:
bors merge

zimbatm:
@JillThornhill can you restore the SUMMARY.md file? It's used as the navigation in the docs.

phaer:
Still failing, I believe the links in SUMMARY.md can't be in a table, that file get's parsed by mdbook - see https://rust-lang.github.io/mdBook/format/summary.html

phaer:
bors merge

PR #116, opened 2023-05-27
Don't claim Linux requirement on source machine

phaer:
As nixos-anywhere is a simple bash script, I'd expect it to run at least pretty much anywhere where nix, bash and curl are supported?

A user mentioned being confused about whether it would be supported to run nixos-anywhere *from* a macOS machine and I believe it should, but personally lack a machine to test it myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,9 @@ data has been migrated.
 ## Prerequisites
 
 - Source Machine:
+
   - Can be any machine with Nix installed, e.g. a NixOS machine.
-  - Should be able to build nix derivations for the target platform. Otherwise 
+  - Should be able to build nix derivations for the target platform. Otherwise
     `--build-on-remote` can be used.
 
 - Target Machine:
--- a/README.md
+++ b/README.md
@@ -67,8 +67,9 @@ data has been migrated.
 ## Prerequisites
 
 - Source Machine:
-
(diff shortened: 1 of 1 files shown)


Lassulus:
Ah it will run on mac if you say --build-on-remote but that raises the memory requirements quite a bit

phaer:
Ah right, that's a good point. It should also work if you already got a remote linux builder defined, which could even be a local nixos vm, right? Sorry if that's a naive question, I am not up to date regarding nix on non-linux systems.

JillThornhill:
In view of Lassulus' comments, I suggest we keep this change, but also add a section in the How To on how to install from a source machine running Mac.

Lassulus:
> It should also work if you already got a remote linux builder defined, which could even be a local nixos vm, right? 

yes that would also work. I still have some code flying around to spin up linux vms on macos, although they dont support cross architectures yet: https://github.com/Lassulus/run-in-linux/

phaer:
> In view of Lassulus' comments, I suggest we keep this change, but also add a section in the How To on how to install from a source machine running Mac.

I just pushed a new suggestion. The section in the how to is a good idea, but I still think the README is inaccurate here and being so early in the README some mac users might not even check the How To before deciding it's not for them.

Lassulus:
ok, now we can finally merge this :D

bors merge

PR #117, opened 2023-05-29
Add in ssh port option

nealfennimore:
Allows for rebuilding a remote host on a non-standard ssh port

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -11,6 +11,8 @@ Options:
   set the flake to install the system from.
 * -i <identity_file>
   selects which SSH private key file to use.
+* -p <ssh_port>
+  set the ssh port to connect with. Defaults to 22.
 * -L, --print-build-logs
   print full build logs
 * -s, --store-paths <disko-script> <nixos-system>
@@ -59,6 +61,7 @@ nix_options=(
 )
 substitute_on_destination=y
 ssh_private_key_file=
+ssh_port=22
 
 declare -A disk_encryption_keys
 declare -a nix_copy_options
@@ -72,6 +75,10 @@ while [[ $# -gt 0 ]]; do
     ;;
   -i)
     ssh_private_key_file=$2
+    shift
+    ;;
+  -p)
+    ssh_port=$2
     shift
     ;;
   -L | --print-build-logs)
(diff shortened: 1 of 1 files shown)


src/nixos-anywhere.sh
```diff
@@ -74,6 +77,10 @@ while [[ $# -gt 0 ]]; do
     ssh_private_key_file=$2
     shift
     ;;
+  -p)
+    ssh_port=$2
```
Kranzes, line 81: It's best to not hardcode 22. Some people have it set to something else in their ssh config.

phaer: Yes, the option should only be passed if explicitly set.

And while we are at it, maybe we could just add `$SSH_ARGS`or so to allow passing `-p` and various `-o` options or would this complicate things here?

nealfennimore: @phaer @Kranzes  I added in the option for passing in ssh port and option via the `-p` and `-o` arg. Let me know if that works

phaer (approved):
Thanks, LGTM!

Kranzes:
i think `-o` and `-p` are too generic for being only for ssh.

nealfennimore:
> i think `-o` and `-p` are too generic for being only for ssh.

How about `--ssh-option` and `--ssh-port`?

Lassulus:
we need to set this port also for the `--build-on-remote` option, I propose adding the port to a separate variable so we can use it in the `--builders` arg if it's actually set. this would also make the filtering out part easier.

Also I'm fine with using -p and -o since ssh is the thing we mostly want to configure. adding the long options additionally would be nice though.

src/nixos-anywhere.sh
```diff
@@ -314,13 +328,21 @@ SSH
 TMPDIR=/root/kexec setsid ${maybe_sudo} /root/kexec/kexec/run
 SSH
 
-  # wait for machine to become unreachable
+  # use the default SSH port to connect at this point
```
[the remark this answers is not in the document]
Mic92: There is also port forwarding i.e. if the machine is behind a router or if you use qemu where the port used on the system is not the same as the port that you need to connect to.

Mic92:
Just saw this PR. I have in the meantime merged this: https://github.com/numtide/nixos-anywhere/pull/130/files
I think we need a combination of both because ssh ports can be both specified in configuration and manually.

src/nixos-anywhere.sh
```diff
@@ -11,6 +11,10 @@ Options:
   set the flake to install the system from.
 * -i <identity_file>
   selects which SSH private key file to use.
+* -p, --ssh-port <ssh_port>
+  set the ssh port to connect with
+* -o, --ssh-option <ssh_option>
```
Mic92, line 16: I think I want to reserve -o for something else, it's not super descriptive on its own.

```suggestion
* --ssh-option <ssh_option>
```

src/nixos-anywhere.sh
```diff
@@ -74,6 +82,14 @@ while [[ $# -gt 0 ]]; do
     ssh_private_key_file=$2
     shift
     ;;
+  -p | --ssh-port)
+    ssh_args+=("-p" "$2")
+    shift
+    ;;
+  -o | --ssh-option)
```
Mic92, line 89:
```suggestion
  --ssh-option)
```

Lassulus:
obsolete by https://github.com/numtide/nixos-anywhere/pull/153

PR #128, opened 2023-06-05
terraform: do not show private key in local-exec logs

jfroche:
While running terraform with and without verbose mode enabled, we don't want the private key content to be shown in the terraform apply output.


zimbatm:
bors merge

PR #131, opened 2023-06-07
Bugfix: Handle parsing ssh hostname on MacOS

adace123:
Ran `nix run github:numtide/nixos-anywhere -- -f ".#$host"  --debug --build-on-remote` on MacOS and got `pubkey='ERROR: failed to retrieve host public key for root@$IP'`.

Looks like the issue is that `ssh -G` on MacOS returns `hostname` instead of `host`. This is causing `ssh_host` to be empty.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -222,7 +222,7 @@ if [[ -n ${ssh_private_key_file-} ]]; then
 fi
 
 ssh_settings=$(ssh -G "${ssh_connection}")
-ssh_host=$(echo "$ssh_settings" | awk '/^host / { print $2 }')
+ssh_host=$(echo "$ssh_settings" | awk '/^host(name?) / { print $2 }')
 ssh_port=$(echo "$ssh_settings" | awk '/^port / { print $2 }')
 
 step Uploading install SSH keys


src/nixos-anywhere.sh
```diff
@@ -222,7 +222,7 @@ if [[ -n ${ssh_private_key_file-} ]]; then
 fi
 
 ssh_settings=$(ssh -G "${ssh_connection}")
-ssh_host=$(echo "$ssh_settings" | awk '/^host / { print $2 }')
+ssh_host=$(echo "$ssh_settings" | awk '/^host(name?) / { print $2 }')
```
Mic92, line 225: Mhm. I cannot reprodue this on a macos machine here:

```
bash-5.2$ ssh -V
OpenSSH_8.6p1, LibreSSL 3.3.6
bash-5.2$  ssh -G foo | awk '/^hostname / { print $2 }' 
foo
bash-5.2$ which ssh
/usr/bin/ssh
```

Mic92: The problem is that there is also a `host` field in the output and it is not the right one:

```
$ ssh -G qemu | head -n 10
host qemu
user joerg
hostname 127.0.0.1
port 2222
addressfamily any
batchmode no
canonicalizefallbacklocal yes
canonicalizehostname false
checkhostip no
compression no
```

adace123: Strange, I don't see the `host` field.
```
sh-5.2$ ssh -G nixos | head -n 10
user aaron
hostname nixos
port 22
addressfamily any
batchmode no
canonicalizefallbacklocal yes
canonicalizehostname false
checkhostip no
compression no
controlmaster false
```

```
sh-5.2$ ssh -V
OpenSSH_9.0p1, LibreSSL 3.3.6
sh-5.2$ ssh -G foo | awk '/^hostname / { print $2 }'
foo
sh-5.2$ which ssh
/usr/bin/ssh
```

```
OS: macOS Big Sur 10.16 22F66 x86_64
Host: MacBookPro16,1
Kernel: 22.5.0
```

Mic92: I think hostname is actually the right field but I am really confused by this diff because we actually already check for this field:

https://github.com/numtide/nixos-anywhere/blob/main/src/nixos-anywhere.sh#L227

Mic92:
Ah. There is a merge conflict. I think we already fixed this issue.

PR #142, opened 2023-06-16
Remove nixos-images/disko override...

phaer:
because nixos-images doesn't have a dependency on disko anymore.


Lassulus:
I wonder if this was ever the case or it's just there by accident?

Lassulus:
bors merge

phaer:
> I wonder if this was ever the case or it's just there by accident?

Not sure either tbh, and didn't bother to check the history. Just verified that it currently isn't the case ;)

PR #146, opened 2023-07-03
move kexec check into kexec codepath

ajs124:
without this, the check gets run and fails even if no kexec will be done e.g. when an installer is booted


Lassulus (approved):

Mic92:
@mergify queue

Mic92:
@mergify rebase

Mic92:
@mergify queue

Mic92:
@mergify refresh

Mic92:
@mergify refresh

Mic92:
@mergify requeue

PR #149, opened 2023-07-09
Changed README & Quickstart

JillThornhill:
Include explanation that configurations will now reside in the flake, not configuration.nix


JillThornhill:
Closed for fix

PR #154, opened 2023-07-13
fix blocking nix-copy-closure

zimbatm:
Pin nix to a version that has a working nix-copy-closure

See https://github.com/NixOS/nix/issues/8480

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -2,7 +2,7 @@
 , openssh
 , gitMinimal
 , rsync
-, nix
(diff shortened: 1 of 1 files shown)


phaer:
Is this about https://github.com/NixOS/nix/issues/8480 ? If so, might be a good idea to put the link in a comment to make it more transparent why it's pinned?

zimbatm:
thanks

phaer:
@mergify queue

PR #157, opened 2023-07-15
export zfs pool after installation

Lassulus:


Mic92:
Does not also require unmounting before Pools can be exported?

Lassulus:
maybe?

phaer:
Did you run in practical problems here? Because I could always zust import nixos-anywhere created zpools on first boot, without any problems?

> Does not also require unmounting before pools can be exported?

zpool export should take care of that, but doesn't hurt ofc ;)

Lassulus:
> Did you run in practical problems here? Because I could always zust import nixos-anywhere created zpools on first boot, without any problems?
> 
> > Does not also require unmounting before pools can be exported?
> 
> zpool export should take care of that, but doesn't hurt ofc ;)

well people ran into it in https://github.com/numtide/nixos-anywhere/issues/156

Mic92:
> Did you run in practical problems here? Because I could always zust import nixos-anywhere created zpools on first boot, without any problems?
> 
> > Does not also require unmounting before pools can be exported?
> 
> zpool export should take care of that, but doesn't hurt ofc ;)

If you have other filesystems mounted on top, than I don't think zpool export will unmount them before hand.

Lassulus:
> > Did you run in practical problems here? Because I could always zust import nixos-anywhere created zpools on first boot, without any problems?
> > > Does not also require unmounting before pools can be exported?
> > 
> > 
> > zpool export should take care of that, but doesn't hurt ofc ;)
> 
> If you have other filesystems mounted on top, than I don't think zpool export will unmount them before hand.

yes we will fail in that case, but that is not a very common one. So in the 99% of cases where people use zfs as the final fs (and not reuse it inside a lvm or mdadm?) we will successfully unmount and export. if not we ignore the error and we are where we were before this PR :)

phaer (approved):

PR #158, opened 2023-07-18
kexec is a path

domenkozar:


Lassulus (approved):

Lassulus:
@mergify queue

phaer:
Is that true?

--kexec sets kexec_url, which is only a path if neither curl nor kexec are available?
if it was, renaming kexec_url to kexec_path would also good to include here?

https://github.com/numtide/nixos-anywhere/blob/e516d378d6a7baea37474347aff94062d5678761/src/nixos-anywhere.sh#L337

domenkozar:
Would `(exec|path)` work?

phaer:
Yes, I do think so. That wouldn't hide the fact that it can be, and the default is, an URL.

PR #160, opened 2023-07-18
skip zpool command if we don't have it installed

Mic92:
for example not all installer will have it.


Lassulus:
but the || : should already take care of that?

Mic92:
> but the || : should already take care of that?

It works but it prints an error message, which looks as if something went wrong to the user.

PR #162, opened 2023-07-20
add --vm-test mode for local vm tests

Lassulus:
depends on https://github.com/nix-community/disko/pull/299


src/nixos-anywhere.sh
```diff
@@ -46,6 +46,8 @@ Options:
   URL of the source Nix store to copy the nixos and disko closure from
 * --build-on-remote
   build the closure on the remote machine instead of locally and copy-closuring it
+* --vm-test
+  run a test locally to debug the disk configuration
```
Mic92, line 50:
```suggestion
  build the system and test the disk configuration inside a VM without installing it to the target.
```

Mic92:
@mergify queue

PR #167, opened 2023-07-31
Default to ssh-ng and and builders-use-substituters

phaer:


phaer:
Ah right, I forgot that ssh-ng still has a few unsolved complexities in relation to other flags, as described for example in https://github.com/NixOS/nix/pull/5010

src/nixos-anywhere.sh
```diff
@@ -46,6 +46,8 @@ Options:
   URL of the source Nix store to copy the nixos and disko closure from
 * --build-on-remote
   build the closure on the remote machine instead of locally and copy-closuring it
+* --legacy-ssh
```
Mic92, line 49: Maybe:

```suggestion
* --builder-protocol
```

Than we can also support other protocols in future?

Mic92:
@

> @phaer any chance you'd want to make it an opt-in feature for now?

Mic92:
Still working on this?

phaer:
Ah no, sorry for leaving this as a draft for so long & thanks for going through stalled PRs :) . I don't plan to set up too many new hosts and don't currently have time for it.

PR #168, opened 2023-07-31
Pass ssh_args to nix copy and nix build SSH invocations

bluepython508:
This is necessary for --post-kexec-ssh-port to be useful in situations involving port forwarding, such as when running on qemu


Mic92:
@mergify rebase

PR #171, opened 2023-08-05
add kexec support for linux-aarch64

Mic92:


Mic92:
Tested on hetzner cloud.

Mic92:
@mergify queue

PR #176, opened 2023-08-13
Don't add trailing newline to keyfile

StillerHarpo:
`pass` will add an trailing newline which renders luks login via prompt useless


phaer:
Nice catch, but as that's mostly an example I think the better solution might be to add a check or just strip trailing newlines for any keyFiles? (i.e. https://github.com/numtide/nixos-anywhere/issues/161#issuecomment-1642332271)

StillerHarpo:
I'm going to fix this in disko directly with https://github.com/nix-community/disko/pull/323. So we can close this without merging after that got merged

phaer:
Thanks, has been merged in disko!

PR #177, opened 2023-08-13
force use of bash to allow for non-posix user shells

adamcstephens:
I use fish, but the current command will likely fail with nushell and others. Make the bash call explicit to avoid failures such as:

```
fish: Unsupported use of '='. In fish, please use 'set is_nixos $(if test -f /etc/os-release && grep -q ID=nixos /etc/os-release; then echo "y"; else echo "n"; fi)'.
```

Ideally this would be handled in `_ssh`, but there are some sporadic args for ssh in some of the calls. A quick perusal looked like they were just connection timeout options, which I'm wondering if we can move into `_ssh`. If this was done, then all calls could pass through bash. For now I've targeted the steps that require posix/bashisms.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -283,7 +283,7 @@ done
 import_facts() {
   local facts filtered_facts
   if ! facts=$(
-    ssh_ -o ConnectTimeout=10 -- <<SSH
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify rebase

adamcstephens:
@mergify rebase

Mic92:
@mergify queue

PR #180, opened 2023-08-16
Fail on trailing new lines in disk-encryption keys

phaer:
draft to see whether with-sudo ci tests fail as expected, currently a bit slow to run locally here ;)


phaer:
ah, hadn't seen those commits. Thanks!

Was wondering whether just stripping newlines could lead to unexpected situations in setups where the keys don't come from an interactive prompt during boot. But we can probably easily fix those should they arise.

PR #198, opened 2023-09-02
extend terraform module to support extra_files_script

Mic92:


threddast:
Would it make sense to trigger this also when a machine gets updated, without having to reinstall from scratch?

Mic92:
Does updating do anything with extra_files?

threddast:
To be clear:

I have a machine up and running:
```
module "deploy" {
  source                 = "github.com/numtide/nixos-anywhere//terraform/all-in-one"
  nixos_system_attr      = "..."
  nixos_partitioner_attr = "..."
  target_host            = ...
  instance_id            = ...
}
```

Then I decide to add the machine to my tailscale network. To do this the machine needs to get an auth-key `tailscale_tailnet_key.test.key` generated by the tailscale provider:

```
resource "tailscale_tailnet_key" "test" {
  depends_on = [tailscale_acl.as_json]
  reusable      = true
  ephemeral     = false
  preauthorized = true
  tags          = ["tag:test"]
}
```

So I change the module accordingly:

```
  extra_files_script     = "${path.module}/deploy-tailscale-authkey.sh"
  extra_environment = {
    TAILSCALE_KEY = tailscale_tailnet_key.test.key
  }
```

and `deploy-tailscale-authkey.sh` does this (adapted it from `nixos-wiki-infra`)
```
...
echo "$TAILSCALE_KEY" >./var/lib/secrets/tailscale
...
```

Now, if I run `tofu apply`, there are no changes detected. I had to `tofu taint module.deploy.module.install.null_resource.nixos-remote`, which worked but after redeploying the server from scratch.
This is the expected behavior, but is there a specific reason for having to redeploying a machine from scratch in these cases?
However, I'm sure there are many other ways to obtain this and I wonder if there is a better way altogether. I'm using sops to deploy other secrets to the machine, but in this case the secret is generated from the terraform tailscale module itself

kstiehl:
Hi @Mic92,

I am currently trying to use this to bootstrap sops.  So this is exactly what I need. 

However tofu OpenTofu in version v1.11.8 will set `path.module` or `path.root` to relative paths. So the lines below break all the paths that come from TF code that were constructed using these expressions.

 https://github.com/nix-community/nixos-anywhere/blob/3243224110cb45887fdc983c377e0b9908333732/terraform/install/run-nixos-anywhere.sh#L67-L70 

I am not 100% sure if OpenTofu always behaved like this, but I can not think of a way to reliably get this to work if the path i use contains `"${path.root}/...."` or `"${path.module}/...."`. These variables are always relative to the module which is evaluating this expression, so changing directories will break the paths constructed like this.

I could make use of `abspath()` but these paths will always change depending on what CI runner, or even what local machine is executing the script and I want to avoid the state file changes that are not required. 

So I am wondering on how this is used on your end, maybe I am missing something very obvious.
If not:

Do you think it could make sense instead of changing the execution context just passing variables to the copy script so that it knows where to put the extra_files. I'd be happy to implement this change if you agree that this makes sense.

PR #199, opened 2023-09-04
readme: clarify VPN scenario + various improvements

DavHau:


Mic92:
Sorry, we just dropped the example as you suggested from the README. Now it's only the quick start. But we can also move the VPN part there.

DavHau:
closing in favor of https://github.com/numtide/nixos-anywhere/pull/211

PR #209, opened 2023-09-11
README: drop example

Mic92:
people don't find the quickstart guide until they are already finished with this one. Also we than have to maintain everything twice


Mic92:
@mergify queue

PR #213, opened 2023-09-14
package: align with nixpkgs version

Mic92:
This is the format we have to use when upstreaming

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


Mic92:
see also https://github.com/NixOS/nixpkgs/pull/255182

Mic92:
@mergify queue

Mic92:
@mergify unqueue

Mic92:
@mergify queue

PR #229, opened 2023-09-29
copy over ssh host keys

tie:
nixos-kexec-installer preserves SSH host keys, so should we, to avoid changing host identity.

See also
- https://github.com/nix-community/nixos-images/blob/c4c73bce65306a1e747684dd0d4bcf0ab2779585/nix/kexec-installer/kexec-run.sh#L42C1-L46C1
- https://github.com/nix-community/nixos-images/blob/c4c73bce65306a1e747684dd0d4bcf0ab2779585/nix/installer.nix#L44-L55

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -29,6 +29,8 @@ Options:
   use another kexec tarball to bootstrap NixOS
 * --post-kexec-ssh-port <ssh_port>
   after kexec is executed, use a custom ssh port to connect. Defaults to 22
+* --copy-host-keys
+  copy over existing /etc/ssh/ssh_host_* host keys to the installation
 * --stop-after-disko
   exit after disko formatting, you can then proceed to install manually or some other way
 * --extra-files <file...>
@@ -117,6 +119,10 @@ while [[ $# -gt 0 ]]; do
     ;;
   --post-kexec-ssh-port)
     post_kexec_ssh_port=$2
+    shift
+    ;;
+  --copy-host-keys)
+    copy_host_keys=y
     shift
     ;;
   --debug)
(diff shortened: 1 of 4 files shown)


src/nixos-anywhere.sh
```diff
@@ -29,6 +29,8 @@ Options:
   use another kexec tarball to bootstrap NixOS
 * --post-kexec-ssh-port <ssh_port>
   after kexec is executed, use a custom ssh port to connect. Defaults to 22
+* --no-copy-host-keys
```
Mic92, line 32: I would like to make this not a default for the following reasons:

- Users might want to copy their own keys in with --extra-files
- I wouldn't trust the original OS to have good ssh keys. Some cloud providers are not shitty and may have the same ssh keys for every OS by accident

So can we invert the logic here and make it opt-in?

tie: > Some cloud providers are not great and may have the same ssh keys for every Recovery OS by accident.

Good point.

> Users might want to copy their own keys in with --extra-files and this option would override it which is surprising (**i.e. when you use agenix and the host key is used to decrypt secrets**)

To be fair, current behavior is also surprising because we lose the host key that is used to decrypt the secrets after installation. We shouldn’t be overwriting existing files in `/mnt` though.

>So can we invert the logic here and make it opt-in?

Agreed, I’ll invert the logic and avoid copying files if the destination already exists.

Now that I think about it, we can have a more generic `--preserve-files` that can be specified multiple times to preserve files from before kexec, and have `--copy-host-keys` be a convenience shortcut to keep SSH host keys. What do you think?

tie: Hm, though that would potentially be non-trivial to implement while keeping all the permissions and directory structure, and I’d like to limit the scope of this PR to copying SSH host keys.

Mic92:
Have you tested this?

I am currently getting this error (see the test I added):

```
installer # + ssh -T -i /tmp/tmp.kBX9n8Yng2/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@installed 'chmod 755 /mnt'
installer # Warning: Permanently added 'installed' (ED25519) to the list of known hosts.
installer # + step Installing NixOS
installer # + echo '### Installing NixOS ###'
installer # ### Installing NixOS ###
installer # + ssh_ bash
installer # + ssh -T -i /tmp/tmp.kBX9n8Yng2/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@installed bash
installer # Warning: Permanently added 'installed' (ED25519) to the list of known hosts.
installer # + export PATH=/run/wrappers/bin:/root/.nix-profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin:/run/current-system/sw/bin
installer # + PATH=/run/wrappers/bin:/root/.nix-profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin:/run/current-system/sw/bin
installer # + mkdir -p /mnt/tmp
installer # + chmod 777 /mnt/tmp
installer # bash: line 18: syntax error near unexpected token `done'
installer # + rm -rf /tmp/tmp.kBX9n8Yng2
```

tie:
Yes, I did test and deploy the initial PR, but didn’t check changes applied after review. There is a minor typo in the shell script, `end` instead of `fi` 🫠

Thank you for adding test case, I’ll update it to verify that `ssh-keyscan` returns the same list of keys.

Lassulus (approved):

Lassulus:
@mergify queue

PR #236, opened 2023-10-19
quickstart fixes

phaer:
trying to fix formatting issues and clarify it a bit, incorporating feedback by @chaoflow at oceansprint

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -21,6 +21,8 @@ You will need:
   performed
 - A disk configuration containing details of the file system that will be
   created on the new server.
+- A target machine, reachable via SSH, with your SSH public key deployed and and
+  the privilege to either login directly as root or to use password-less sudo.
 
 **nixos-anywhere** doesn’t need to be installed. You can run it directly from
 [Numtide's repository on Github.](https://github.com/numtide/nixos-anywhere)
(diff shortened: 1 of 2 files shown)


chaoflow:
@phaer https://github.com/nix-community/nixos-anywhere/blob/main/docs/quickstart.md#steps-required-to-run-nixos-anywhere --- Step 2, ``nix flake init`` seems to not be needed, I suggest to instead explain to copy all files from the example repo, or use flake template as discussed today.

cc @Mic92

phaer:
@chaoflow Ah right, I was a bit lazy yesterday and sleep-deprived enough that I wasn't sure whether my changes would be a net-positive in terms of readability. Still not totally recovered, but that might be good enough?

phaer:
Didn't really get what you meant here: an scp command? L

phaer:
I thought that was almo

docs/quickstart.md
```diff
@@ -70,11 +63,12 @@ below.
   ```
 
   This creates the necessary configuration files under `/mnt/etc/nixos/`, which
-  you can then customize as needed.
+  you can then customize as needed and copy them to your local machine in order
```
Mic92, line 66: Ah, you kind of mentioning this here. I missed that sentence on my phone.

Mic92:
@mergify queue

docs/quickstart.md
phaer: Ah, just saw that I somehow posted a butchered response from my phone, hope that wasn't confusing

PR #238, opened 2023-10-26
nixos-anywhere.sh: recursive umount before zfs export

PaulGrandperrin:
fixes zpool export error when some FS are mounted under some ZFS mount.

When ZFS is used as the root filesystem, other FS will be mounted under it, like at `/mnt/boot`.
`zpool export -a` will try to unmount `/mnt` and fail.
This will then prevent the installed system from booting without forcing the import with `zfs_force` in the kernel command line.


phaer:
Ah sorry, hadn't seen that one and just merged https://github.com/nix-community/nixos-anywhere/pull/239/files

PR #240, opened 2023-10-28
only fetch pubkey in case we want to build on remote

Mic92:
This is a bit faster and also helps to mitigated cases where additionally options that are not respected by ssh-keyscan (i.e. JumpHost) are used


phaer (approved):
LGTM, was wondering about usage but pubkey is only ever used if remote builds are active

@mergify queue

phaer:
@mergify queue

PR #234, opened 2023-10-15
flake.lock: Update

Mic92:
Automatic buildbot update

PR #243, opened 2023-11-02
Prefer system rsync

### LINKED ISSUE #242: Using --extra-files on a non-NixOS source machine causes crash with newer lockfile
## Steps to reproduce

Source machine:

```
$ nix run nixpkgs#nix-info -- -m
 - system: `"aarch64-linux"`
 - host os: `Linux 6.2.0-36-generic, Ubuntu, 22.04.3 LTS (Jammy Jellyfish), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

Set up target machine with the 23.05 installer.

* Clone repo flake to a local directory.
* Run `nix flake lock --update-input nixpkgs`.
* Do a nixos-anywhere run with `--extra-files`; e.g.

```
nix run ~/nixos-anywhere -- --debug --flake ./src/etc/nixos#nixos --extra-files ./src root@TARGET
```

## Expected behaviour

Normal install completed w/files copied.

## Actual behaviour

rsync crashes when uploading files. Abridged output:

```
### Copying extra files ###                                                                                                            
+ rsync -rlpv -FF -e 'ssh -i "/tmp/tmp.*****"/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ' ./src/ root@192.168.xx.xx:/mnt/                                                                                                                    
Warning: Permanently added '192.168.xx.xx' (ED25519) to the list of known hosts.                                                       
sending incremental file list                                                                                                          
*** buffer overflow detected ***: terminated                                                                                           
/nix/store/xp5ql5d6b2n6kiw4aqz6alvxg1vp5psy-nixos-anywhere-1.0.0/bin/.nixos-anywhere-wrapped: line 449: ... Aborted (core dumped) ...  
```

## Additional notes

I initially found this with the nixpkgs package; I guess something has recently changed that has caused mixing system SSH with Nix rsync to crash. Adding rsync to the suffix list that was added in #66 seems to fix the issue.

vancluever:
This sets up the package derivation to prefer the system rsync in the same way the system SSH is preferred, to prevent issues that may arise (e.g. crashes) that can arise from mixing the system SSH with the derivation-provided rsync.

Fixes #242.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -18,7 +18,6 @@
 let
   runtimeDeps = [
     gitMinimal # for git flakes
-    rsync
     # pinned because nix-copy-closure hangs if ControlPath provided for SSH: https://github.com/NixOS/nix/issues/8480
     (if lib.versionAtLeast nix.version "2.16" then nix else nixVersions.nix_2_16)
     coreutils
@@ -39,13 +38,15 @@ stdenv.mkDerivation {
 
     # We prefer the system's openssh over our own, since it might come with features not present in ours:
     # https://github.com/numtide/nixos-anywhere/issues/62
+    #
+    # We also prefer system rsync to prevent crashes between rsync and ssh.
     wrapProgram $out/bin/nixos-anywhere \
-      --prefix PATH : ${lib.makeBinPath runtimeDeps} --suffix PATH : ${lib.makeBinPath [ openssh ]}
+      --prefix PATH : ${lib.makeBinPath runtimeDeps} --suffix PATH : ${lib.makeBinPath [ openssh rsync ]}
   '';
 
   # Dependencies for our devshell
   passthru.devShell = mkShellNoCC {
-    packages = runtimeDeps ++ [ openssh terraform-docs ];
+    packages = runtimeDeps ++ [ openssh rsync terraform-docs ];
   };
 
   meta = with lib; {


vancluever:
PS: Not putting the updated lockfile in this one, but I'd imagine that this should also be shipped to nixpkgs even if it doesn't result in an actual version bump.

Mic92:
@mergify queue

Mic92:
Are we sure that this has something to do with rsync/ssh not being compatible and not with this regression? https://github.com/NixOS/nixpkgs/issues/265014

nyawox:
Seems like `nix profile install nixpkgs/nixos-23.05#rsync` on local machine running unstable works as temporary easy workaround
Installed successful now

vancluever:
Hmm, yeah interesting! Admittedly for sure my only real need for rsync recently has been for nixos-anywhere, so if this is indeed the issue this can just be reverted - that'd fix the nixpkgs invocation too and the flake lockfile here could just be fast-forwarded when the build is fixed in nixpkgs/nixos-unstable.

Mic92:
I am currently working on a proper fix: https://github.com/nix-community/nixos-anywhere/pull/234

PR #245, opened 2023-11-06
README: add missing line

zowoq:
https://github.com/nix-community/disko/blob/master/README.md?plain=1#L135

Matches the disko readme.


phaer:
@mergify queue

PR #247, opened 2023-11-08
support IPv6 address when using --extra-files

### LINKED ISSUE #246: Copying extra files fails when using IPv6 address
I'm using the `1.1.0` release.

Here's my invocation:

```console
nix run github:numtide/nixos-anywhere/1.1.0 -- \
  root@2600:4040:2094:c800:ca7f:54ff:fed0:db4 \
  -i ~/.ssh/id_ed25519 \
  --flake .#nixos-build01 \
  --kexec https://gh-v6.com/nix-community/nixos-images/releases/download/nixos-unstable/nixos-kexec-installer-noninteractive-x86_64-linux.tar.gz \
  --print-build-logs \
  --debug \
  --extra-files /Volumes/nixos-build01
```

My configuration files are here: <https://github.com/connorbaker/nixos-configs>.

The script appears to work up to the point where it tries to copy the extra files over:

```console
+ step Copying extra files
+ echo '### Copying extra files ###'
### Copying extra files ###
+ rsync -rlpv -FF -e 'ssh -i "/tmp/tmp.1y1HPByasE"/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ' /Volumes/nixos-build01/ root@2600:4040:2094:c800:ca7f:54ff:fed0:db4:/mnt/
ssh: connect to host 0.0.10.40 port 22: Connection timed out
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(231) [sender=3.2.7]
+ rm -rf /tmp/tmp.1y1HPByasE
```

At that point it errors out. However, I can ssh into the machine and see that it is still running, albeit without the extra files copied over.

Here are the relevant lines in the script:

https://github.com/nix-community/nixos-anywhere/blob/7973a36f24dd2bc4258152ff6fb6ea63670c5a6c/src/nixos-anywhere.sh#L450-L455

Looking at the logs pertaining to `rsync`, we see `ssh: connect to host 0.0.10.40 port 22: Connection timed out`. That's not the IP address that was specified... but if you convert the first hextet of the IPv6 address, 2600, to IPv4 octets you do indeed get `10.40` ($`2600 = 10*256^1 + 40*256^0`$).

I think that because `rsync` uses a colon (`:`) to delimit the remote from the destination path, it's being tripped up by the colons in the IPv6 address.

Here's a similar post which describes a solution: https://serverfault.com/questions/420158/how-do-i-use-rsync-with-an-ipv6-address-via-ssh. In short, use brackets (`[]`) to enclose the IPv6 address.

ConnorBaker:
Fixes #246.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -447,11 +447,21 @@ if [[ -n ${extra_files-} ]]; then
   if [[ -d $extra_files ]]; then
     extra_files="$extra_files/"
   fi
+  # If `ssh_host` is an IPv6 address, we need to wrap it in brackets.
+  if [[ $ssh_host =~ ^[0-9a-fA-F:]+$ ]]; then
+    # TODO: If the user changes from `root` to something else, this will fail.
+    rsync_destination="root@[${ssh_host}]:/mnt/"
+  else
+    # NOTE: Yes, `ssh_connection` is exactly `root@${ssh_host}`, but we want to show
+    # that IPv6 and IPv4 addresses differ only in whether they are wrapped in brackets.
+    # Indirection through another variable like `ssh_connection` obscures this.
+    rsync_destination="root@${ssh_host}:/mnt/"
+  fi
   step Copying extra files
   rsync -rlpv -FF \
     -e "ssh -i \"$ssh_key_dir\"/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ${ssh_args[*]}" \
     "$extra_files" \
-    "${ssh_connection}:/mnt/"
+    "$rsync_destination"
   ssh_ "chmod 755 /mnt" # rsync also changes permissions of /mnt
 fi
 


ConnorBaker:
@Mic92 would you have an opportunity to review this?

src/nixos-anywhere.sh
```diff
@@ -447,11 +447,21 @@ if [[ -n ${extra_files-} ]]; then
   if [[ -d $extra_files ]]; then
     extra_files="$extra_files/"
   fi
+  # If `ssh_host` is an IPv6 address, we need to wrap it in brackets.
+  if [[ $ssh_host =~ ^[0-9a-fA-F:]+$ ]]; then
```
phaer, line 451: This regex could result in quite a few false positives IMO.

phaer (changes requested):
Are you sure this is needed? I think rsync supports square brackets around ipv4 addresses and hostnames as well for exactly this reason and quick testing seems to support that hypothesis, so i.e.:

    rsync_destination="root@[${ssh_host}]:/mnt/"

should work the same?

Mic92:
Lack of response. Please re-open if you plan to continue working on this.

PR #252, opened 2023-11-22
Replace `numtide` with `nix-community`

drupol:


phaer:
Thanks, looks good but please run treefmt i.e. `nix fmt` should do :)

phaer:
@mergify queue

PR #255, opened 2023-11-27
Fix url for example repo

vakili:
Takes care of instances not handled by https://github.com/nix-community/nixos-anywhere/pull/253


phaer:
@mergify queue

PR #258, opened 2023-11-30
fix: Set BatchMode for ssh-copy-id

dantefromhell:
`ssh-copy-id` can hang indefinitely at the SSH password prompt if Server is configured to allow KeyboardAuthentication and PubKeyAuth fails.

SSH BatchMode instructs the ssh client to disable all interactive user prompts like password and host key confirmation.

Using the terraform module without `BatchMode` can result in a terraform apply run hanging "forever" at the password prompt.  With BatchMode enabled the terraform apply run fails correctly.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -288,6 +288,7 @@ step Uploading install SSH keys
 until
   ssh-copy-id \
     -i "$ssh_key_dir"/nixos-anywhere.pub \
+    -o BatchMode=yes \
     -o ConnectTimeout=10 \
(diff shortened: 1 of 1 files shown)


dantefromhell:
Digging through the code a bit more I suspect this could happen during all steps of `nixos-anywhere` that use SSH.

I guess the question is if running `nixos-anywhere` from the CLI should support the use-case of entering the password and this issue would be better fixed by the terraform module setting `BatchMode=yes` via the `--ssh-option` parameter?

Please advice.

PR #261, opened 2023-12-10
nixos 23.05 -> 23.11

phaer:


flake.nix
```diff
@@ -7,10 +7,10 @@
 
     # used for testing
     disko = { url = "github:nix-community/disko/master"; inputs.nixpkgs.follows = "nixpkgs"; };
-    nixos-2305.url = "github:NixOS/nixpkgs/release-23.05";
+    nixos-2311.url = "github:NixOS/nixpkgs/nixos-23.11";
```
phaer, line 10: Is there a reason we've used 'release-*' here before, @Mic92 ?

Mic92: Unawareness?

PR #265, opened 2023-12-27
[feat] remove -t parameter and replace it with a check of whether it's tty

MangoIV:
I haven't tested the manual version yet...

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -20,9 +20,6 @@ Options:
 * -s, --store-paths <disko-script> <nixos-system>
   set the store paths to the disko-script and nixos-system directly
   if this is give, flake is not needed
-* -t --tty
-  Force pseudo-terminal allocation in SSH sessions. Use this when you expect e.g.
-  to be asked for password entry during LUKS configuration.
 * --no-reboot
   do not reboot after installation, allowing further customization of the target installation.
 * --kexec <path>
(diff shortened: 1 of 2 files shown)


Mic92:
Looks good otherwise.

MangoIV:
Yeah I’ll probably get the chance to try it out soon, I’ll report back then. :)

MangoIV:
it works...

src/nixos-anywhere.sh
```diff
@@ -180,6 +174,10 @@ if [[ ${print_build_logs-n} == "y" ]]; then
   nix_options+=("-L")
 fi
 
+if [ -t 0 ]; then
+  ssh_tty_param="-t"
+fi
```
zimbatm, line 179: Move this to line 71 so it becomes the detected default.

src/nixos-anywhere.sh
```diff
@@ -106,9 +103,6 @@ while [[ $# -gt 0 ]]; do
     shift
     shift
     ;;
-  -t | --tty)
-    ssh_tty_param="-t"
-    ;;
```
zimbatm, line 111: This is useful for tools that might want to wrap up this program and interact with it as if they had a TTY.

MangoIV: I would leave it in if @Mic92 and @RaitoBezarius don't have anything against it.

Mic92: @zimbatm according to @Lassulus ssh blocks if the input is a pipe and not a tty.

zimbatm: Ok, I defer back to Lassulus

Mic92:
@mergify queue

PR #266, opened 2023-12-27
terraform/install: support `build_on_remote`

Enzime:
Remove the trusted user requirement for `--build-on-remote` by using `--store` instead of `--builders`

This is very useful for cross architecture deploys especially without any remote builders configured


Mic92:
@mergify queue

PR #273, opened 2024-01-15
Fix container detection

c-14:
When systemd-detect-virt exists and we aren't in a container, is_container contains `y none none` instead of `none`.
This causes nixos-anywhere to warn `WARNING: This script does not support running from a 'y' container. kexec will likely not work` even though we aren't in a container.

1. has outputs y/n to stdout which we need to remove
2. `systemd-detect-virt --container` returns non-zero when we aren't in a container, use if then else instead of && || to prevent duplicating `none` in is_container variable

Tested running against a stock debian 12 with bash as login shell


c-14:
Converted to draft because has() can never (reasonably) return non-zero as it returns the exit code of echo itself and as such `has systemd-detect-virt` will always return truthy. This is also an issue in the original commit which I'll fix (probably tomorrow) and update this PR.

c-14:
Switched to checking the output of has() instead of its return value. Not sure if this is conceptually cleaner than say using command -v directly but it fits the style of the other facts

Mic92:
@mergify queue

PR #281, opened 2024-02-12
removed duplicate "/tmp/disk-1.key /tmp/disk-1.key"

we-do-it-lu:


scvalex:
Isn't this change actually wrong? 

The docs say both `<remote_path>` and `<local_path>` are necessary.

```
* --disk-encryption-keys <remote_path> <local_path>
  copy the contents of the file or pipe in local_path to remote_path in the installer environment,
  after kexec but before installation. Can be repeated.
```

The code also doesn't seem to have any support for eliding the second path if it's identical: https://github.com/nix-community/nixos-anywhere/blob/b6abac7b0e2405b1ee4da44b6130c4571b2ed8bb/src/nixos-anywhere.sh#L137

Mic92:
Yeah it is.

PR #293, opened 2024-03-23
Switch to root user if non-root user is used for nixos installer

### LINKED ISSUE #279: no-os install requires root login
I am installing NixOS with a booted installer CD this document mentions it can be done as nixos user that is present on the installer cd.

https://github.com/nix-community/nixos-anywhere/blob/4c94cecf3dd551adf1359fb06aa926330f44e5a6/docs/howtos/no-os.md?plain=1#L22

but I received these errors:

```
+ wipefs --all -f /dev/xvda1
wipefs: error: /dev/xvda1: probing initialization failed: Permission denied
+ wipefs --all -f /dev/xvda2
wipefs: error: /dev/xvda2: probing initialization failed: Permission denied
+ wipefs --all -f /dev/xvda3
wipefs: error: /dev/xvda3: probing initialization failed: Permission denied
++ zdb -l /dev/xvda
++ sed -nr 's/ +name: '\''(.*)'\''/\1/p'
+ zpool=
+ [[ -n '' ]]
+ unset zpool
++ lsblk /dev/xvda -l -p -o type,name
++ awk 'match($1,"raid.*") {print $2}'
+ md_dev=
+ [[ -n '' ]]
+ wipefs --all -f /dev/xvda
wipefs: error: /dev/xvda: probing initialization failed: Permission denied
+ dd if=/dev/zero of=/dev/xvda bs=440 count=1
dd: failed to open '/dev/xvda': Permission denied
++ mktemp -d
+ disko_devices_dir=/tmp/tmp.FR9K6lAylU
+ trap 'rm -rf "$disko_devices_dir"' EXIT
+ mkdir -p /tmp/tmp.FR9K6lAylU
+ device=/dev/xvda
+ imageSize=2G
+ name=disk1
+ type=disk
+ device=/dev/xvda
+ efiGptPartitionFirst=1
+ type=gpt
+ sgdisk --align-end --new=1:0:+1M --change-name=1:disk-disk1-space --typecode=1:EF02 /dev/xvda
Problem opening /dev/xvda for reading! Error is 13.
You must run this program as root or use sudo!
```

I'd be happy to update the no-os doc to reflect that root is required.  Or if you'd like to double check that sudo is being called in the appropriate places in the code.

Let me know

nbdd0121:
Fix #279

Currently nixos-anywhere works with non-root users on non-NixOS installer/recovery images since it will kexec into a NixOS installer and proceed as root. But it doesn't work for NixOS installer if non-root user is used. Non-root user can be desirable if a preconfigured image is used both as Live CD and installer.

This PR allows non-root configuration by copying ~/.ssh/authorized_keys to /root/.ssh/authorized_keys and proceed as root (this should be fine regardless how user logs into the non-root user, since the nixos-anywhere.pub is what matters).

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -282,6 +282,7 @@ if [[ -n ${ssh_private_key_file-} ]]; then
 fi
 
 ssh_settings=$(ssh "${ssh_args[@]}" -G "${ssh_connection}")
+ssh_user=$(echo "$ssh_settings" | awk '/^user / { print $2 }')
 ssh_host=$(echo "$ssh_settings" | awk '/^hostname / { print $2 }')
 ssh_port=$(echo "$ssh_settings" | awk '/^port / { print $2 }')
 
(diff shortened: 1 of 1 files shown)


ibizaman:
FWIW I confirm this works. I just tested this locally using the NixOS USB installer.

robbins:
Why is the root user required instead of just elevating privileges with `sudo`?

PR #295, opened 2024-03-26
migrate rsync code to scp

Lassulus:


src/nixos-anywhere.sh
```diff
@@ -466,11 +466,8 @@ if [[ -n ${extra_files-} ]]; then
     extra_files="$extra_files/"
   fi
   step Copying extra files
-  rsync -rlpv -FF \
```
Mic92, line 469: Things to check is if it actually restores permissions/owner correctly.

Mic92: Looks like tests are not happy yet.

0x4A6F: There is also a catch, that the ssh-opt and scp-opt differ (i.e. port).

Mic92:
Closing in favor of https://github.com/nix-community/nixos-anywhere/pull/316

PR #297, opened 2024-04-05
fix bashism

zimbatm:
[[ ]] is not available in POSIX sh, and test -o is not well defined either.

Move the facts-gathering script out so it can be tested with shellcheck.


Mic92:
@mergify queue

PR #302, opened 2024-04-11
Allow installs through terraform to work on different ports

Drakonis:
should fix issues with installing to ports other than the default


otavio (approved):

Mic92:
@mergify queue

PR #304, opened 2024-04-21
Detect IPv6 automatically, Rewrite kexec url

mkg20001:
This uses ip r g 1.1.1.1 to test if there's any route to an IPv4 internet host

If that fails and the url is a github.com one it will change github.com to gh-v6.com,
as specified in the documentation

This saves the user from an extra manual step and makes nixos-anywhere better

The fallback behaviour if the ip binary cannot be found is to just assume ipv4 connectivity. I can invert that if wanted.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -18,7 +18,7 @@ Options:
 * -L, --print-build-logs
   print full build logs
 * --env-password
-  set a password used by ssh-copy-id, the password should be set by 
+  set a password used by ssh-copy-id, the password should be set by
   the environment variable SSH_PASS
 * -s, --store-paths <disko-script> <nixos-system>
   set the store paths to the disko-script and nixos-system directly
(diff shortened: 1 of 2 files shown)


src/nixos-anywhere.sh
```diff
@@ -316,6 +316,7 @@ is_kexec=\$(if test -f /etc/is_kexec; then echo "y"; else echo "n"; fi)
 is_nixos=\$is_nixos
 is_installer=\$(if [[ "\$is_nixos" == "y" ]] && grep -q VARIANT_ID=installer /etc/os-release; then echo "y"; else echo "n"; fi)
 is_container=\$(if [[ "\$(has systemd-detect-virt)" == "y" ]]; then systemd-detect-virt --container; else echo "none"; fi)
+has_ipv4=\$(if [[ "\$(has ip)" == "n" ]] || ip r g 1.1.1.1 >/dev/null 2>/dev/null; then echo "y"; else echo "n"; fi)
```
Mic92, line 319: I would also detect if ipv6 is routable.

```suggestion
has_ipv6_only=\$(if [[ "\$(has ip)" == "n" ]] || ip r g 1.1.1.1 && ip r g 2606:4700:4700::1111 >/dev/null 2>/dev/null; then echo "y"; else echo "n"; fi)
```

src/nixos-anywhere.sh
```diff
@@ -381,6 +382,11 @@ $maybe_sudo rm -rf /root/kexec
 $maybe_sudo mkdir -p /root/kexec
 SSH
 
+  # no way to reach global ipv4 destinations, use gh-v6.com automatically if github url
+  if [[ ${has_ipv4-n} == "n" ]] && [[ $kexec_url == "https://github.com/"* ]]; then
```
Mic92, line 386:
```suggestion
  if [[ ${has_ipv6_only-n} == "n" ]] && [[ $kexec_url == "https://github.com/"* ]]; then
```

Mic92:
cc @zimbatm because he is paying for the proxy.

Mic92:
This pull request also needs to be rebased once https://github.com/nix-community/nixos-anywhere/pull/297 is merged.

Mic92:
@mergify queue

PR #305, opened 2024-04-23
feat: add copy-password

a-kenji:
Allow setting the bootstrap password on invocation of `nixos-anywhere`.

This makes use of the `sshpass` program, to provide `ssh-copy-id` with the password.

The `runtimeDeps` change in the following way:

sshpass: ∅ → 1.10, +29.5 KiB

Improves the usage especially together with the iso image installers of `nixos-images`.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -17,6 +17,9 @@ Options:
   set an ssh option
 * -L, --print-build-logs
   print full build logs
+* --env-password
+  set a password used by ssh-copy-id, the password should be set by 
+  the environment variable SSH_PASS
 * -s, --store-paths <disko-script> <nixos-system>
   set the store paths to the disko-script and nixos-system directly
   if this is give, flake is not needed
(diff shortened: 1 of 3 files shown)


src/nixos-anywhere.sh
```diff
@@ -17,6 +17,9 @@ Options:
   set an ssh option
 * -L, --print-build-logs
   print full build logs
+* --copy-password
```
Lassulus, line 20: this sounds a bit confusing? maybe we should call that more descriptive? maybe --env-password or --password-via-env?

a-kenji: Yes, that is fair.
I also thought about it prior, but wanted to highlight that this password is used only for  `ssh-copy-id`.

I would also be fine with
```
--env-password
```
if you think it is enough to specify in the help, that it is used for `ssh-copy-id`.

malteneuss:
So when we create a new virtual machine that has a predefined user password (i'm planning of trying out Hetzner), instead of logging-in manually and adding our local ssh key into the VM's `authorized_keys` file to be able to login, we can just use something along the lines of

```bash
SSH_PASS=<vm-password> nixos-anywhere --env-password <user>@<ip>...
```

directly?

That would be/is a great quality of life improvement!

a-kenji:
Yes, that was the motivation, though I haven't tested it yet with hetzner.
I have tested it with the `nixos-images` so far.
I would be happy to get feedback about it.

PR #310, opened 2024-05-07
Add nix option in terraform and fix run-nixos-anywhere.sh

misumisumi:
The current terraform module does not accept nix options.
Therefore, it may not work properly when needing `--options pure-eval false` and so on.
I add `nix_options` to the terraform module to address this issue.

Also, I fixed an issue where `$tmpdir/keys` could not be created again when multiple `disk_encryption_key_scripts` were specified in `run-nixos-anywhere.sh`, resulting in an error.


misumisumi:
> Also, I fixed an issue where $tmpdir/keys could not be created again when multiple disk_encryption_key_scripts were specified in run-nixos-anywhere.sh, resulting in an error.

The example shown in [All-In-One](https://github.com/nix-community/nixos-anywhere/blob/main/terraform/all-in-one.md) seems wrong.
`decrypt-zfs-key.sh` is shown to redirect, but even within the script I am trying to create a file with the redirect.
As shown in `./decrypt.sh > $tmpdir/key/0 (./decrypt.sh like echo "test" > key.txt)`, there is a double redirect, but is it as intended?

terraform/nix-build/nix-build.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 set -efu
 
-declare file attribute
-eval "$(jq -r '@sh "attribute=\(.attribute) file=\(.file)"')"
+declare file attribute nix_options
+eval "$(jq -r '@sh "attribute=\(.attribute) file=\(.file) nix_options=\(.nix_options)"')"
 if [[ -n ${file-} ]] && [[ -e ${file-} ]]; then
-  out=$(nix build --no-link --json -f "$file" "$attribute")
+  out=$(nix build --no-link --json $(echo "$nix_options") -f "$file" "$attribute")
```
Mic92, line 7: Shellcheck doesn't like this line.
Does jq handles the quoting here correctly?

Mic92:
> > Also, I fixed an issue where $tmpdir/keys could not be created again when multiple disk_encryption_key_scripts were specified in run-nixos-anywhere.sh, resulting in an error.
> 
> The example shown in [All-In-One](https://github.com/nix-community/nixos-anywhere/blob/main/terraform/all-in-one.md) seems wrong. `decrypt-zfs-key.sh` is shown to redirect, but even within the script I am trying to create a file with the redirect. As shown in `./decrypt.sh > $tmpdir/key/0 (./decrypt.sh like echo "test" > key.txt)`, there is a double redirect, but is it as intended?

Fixed. I checked again with the orignal machine where I took this from.

terraform/nix-build/main.tf
```diff
@@ -1,8 +1,12 @@
+locals {
+  nix_options = var.nix_options == null ? "" : join(" ", [for k, v in var.nix_options :  "--option ${k} ${v}"])
```
Mic92, line 2: I guess we already give up on quoting here already.

Mic92: Could we not dump this as a json object here?
And than on the shell script side, we use jq to get the data back?
Than we can build up some array in bash instead with all quoting handled correctly.

misumisumi: I changed it to json format and handled it on the shell side.

misumisumi:
@Mic92 
Sorry for the late reply and work.
I have corrected the points you pointed out, so please check.,

terraform/nix-build/main.tf
```diff
@@ -1,8 +1,12 @@
+locals {
+nix_options = "{\"options\": {${join(", ", [for k, v in var.nix_options :  "\"${k}\": \"${v}\""])} } }"
+}
```
Mic92, line 3: Terraform has actually a nice function for this:

```suggestion
locals {
  nix_options = jsonencode({
    options = { for k, v in var.nix_options : k => v }
  })
}
```

misumisumi:
Thank you for the correction.

misumisumi:
On a separate matter, there is no `ignore_systemd_errors` option in all-in-one, is this intentional?
If not, is it okay to add it? or should I separate PR?

Mic92:
@mergify queue

Mic92:
> On a separate matter, there is no `ignore_systemd_errors` option in all-in-one, is this intentional? If not, is it okay to add it? or should I separate PR?

Probably just an oversight, we can add it there as well.

PR #315, opened 2024-05-09
fix(README): fix ram requirements without kexec

Prince213:
It seems that RAM requirements only apply when using kexec.

References (conflicting docs):
- https://github.com/nix-community/nixos-anywhere/blob/main/docs/requirements.md
- https://github.com/nix-community/nixos-anywhere/blob/main/docs/howtos/no-os.md


Mic92:
@mergify queue

Prince213:
Thanks!

PR #332, opened 2024-06-07
Add new parameter "--kexec-extra-flags"

johbo:
This allows to pass in special flags like "--no-sync" into the call to "kexec".

This does depend on a PR to `nixos-images`:
- https://github.com/nix-community/nixos-images/pull/228

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1717637172,
-        "narHash": "sha256-geTO9YL1V9zYGxupuZvKFWxh6II9sH2bjI1dmEVhVYQ=",
+        "lastModified": 1718008439,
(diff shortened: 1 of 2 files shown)


Mic92:
`--kexec-extra-flags` has been merged now.

Mic92:
Can you do an nix-flake update, so we actually use the new code in the test?

johbo:
> Can you do an nix-flake update, so we actually use the new code in the test?

Did the update, not fully sure about it though. I've in a first commit only done the `nixos-images` which also did apply a change caused probably by an older refactoring, a rename from `nixos-2311` to `nixos-stable` ?

Added a second commit which updates all dependencies. I am not sure which flow prefer, e.g. update them all vs. just update the minimal set in this PR. 😬

johbo:
Oh wait, I missed a rebase, will update one more time ⏲️

Mic92:
@mergify queue

Issue #278, opened 2024-02-02
Add option to prevent unmounting and exporting ZFS pool


TECHNOFAB11:
I just tried to use nixos-anywhere for a new server and noticed it now unmounts and exports the ZFS pool.
I need these to stay mounted though so that I can copy over the SSH keys for SOPS.
An option to prevent that would be nice, like no-reboot, maybe no-u(n)mount?

sedlund:
I have the same use case.  It's actually in disko here:

https://github.com/nix-community/disko/blob/d07de570ba05cec2807d058daaa044f6955720c7/lib/default.nix#L440

haven't dug in yet to see how it could be bypassed.

TECHNOFAB11:
In the nixos-anywhere.sh at the end it also exports all ZFS pool, I don't think the Disko destroy script is currently used for that as that presumably deletes partitions?

sedlund:
yeah i think I jumped the gun before I had to head out.  I looked at that script and it tests for zpool command and does that recurive unmount.

I'm on btrfs though so I dont think that's my issue.  

I decided I'm going to pre-provision ssh keys and put them in sops before running nixos-anywhere with `--extra-files` to copy them onto the host like this: 
 https://github.com/nix-community/nixos-anywhere/blob/4c94cecf3dd551adf1359fb06aa926330f44e5a6/docs/howtos/secrets.md#example-decrypting-an-openssh-host-key-with-pass

then the initial build will work and I won't have to come back after to update my secrets, rebuild, and redeploy.

Mic92:
Yes `--extra-files` should be preferably used to bootstrap secrets. Would be cool if someone could make this more clear in our quickstart if this is not already the case - Maybe even some sops-nix section...

fin444:
> `--extra-files` should be preferably used to bootstrap secrets

Isn't it insecure to create secrets on one machine and copy them to another?

sedlund:
> Isn't it insecure to create secrets on one machine and copy them to another?

Not if done correctly.  If secrets are clear text and written to files, they should be written correctly permissioned to tmpfs (ram), or encrypted storage, network copied over a secure transport (ssh), and stored similarly on the remote.

[...]

PR #342, opened 2024-06-29
don't attempt to export zpool if there is none

fin444:
Currently, `/mnt` will be unmounted and pools will be exported if `zpool` is in the `PATH`. However, that doesn't mean there are actually pools to export.

Partial fix for https://github.com/nix-community/nixos-anywhere/issues/278


Mic92:
@mergify queue

PR #344, opened 2024-07-01
Fix two issues caused by recent changes

r-vdp:
- **Fix typo that causes non-flake builds to fail**
- **Fix error caused by nix_options defaulting to null in all-in-one**


zimbatm (approved):
LGTM

PR #354, opened 2024-08-05
get-facts: Make os-release parser more generic

dasJ:
This adds support for quoted values (which are allowed by the file format) and also for `ID_LIKE` for systems that fork NixOS.


0x4A6F (approved):
Tested with both native NixOS ISO and ISO with custom ID.

Doesn't unconditionally enter `kexec` stage anymore.

PR #356, opened 2024-08-08
Timeout: add seperate flags for reboot and ssh timeout

hsjobeki:
cc: @Lassulus since you added the infinite loops; Can you check if this works for you?


Lassulus:
it seems the script aborts after the first ssh error if the retry_limits are not defined

Lassulus (approved):
lgtm, but not tested

phaer:
Logic LGTM, but as written `--reboot-retry-limit` doesn't seem to have to do anything with the actual reboot (into the installed system, at the very end), but just retries connecting after kexec. While kexec could be considered a form of reboot, I think this naming could confuse users and would suggest something like `--kexec-connection-retries` or so

PR #358, opened 2024-08-13
Add option --ssh-store-settings

aabccd021:
This flag allows users to run nixos-anywhere with [extra settings for SSH store](https://nix.dev/manual/nix/2.22/store/types/ssh-store), by appending it to ssh store url.

If a user has fast connection between client and target, but slow internet connection on target, the user can use `--no-substitute-on-destination` to copy closures from client instead of downloading it from internet. 

The copy speed (nix-copy to ssh store) can be greatly improved by enabling compression via `--ssh-store-settings "compress=true"`.

Existing url parameters (especially the special characters) need to be encoded to avoid unexpected results. Otherwise we will get something like `root` evaluates to `/mnt?compression=true`.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -638,7 +638,7 @@ runDisko() {
     # If we don't use ssh-ng here, we get `error: operation 'getFSAccessor' is not supported by store`
     diskoScript=$(
       nixBuild "${flake}#${flakeAttr}.system.build.${diskoMode}Script" \
-        --eval-store auto --store "ssh-ng://$sshConnection?ssh-key=$sshKeyDir/nixos-anywhere"
+        --eval-store auto --store "ssh-ng://$sshConnection?ssh-key=$sshKeyDir%2Fnixos-anywhere"
     )
   fi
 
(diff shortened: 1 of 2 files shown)


aabccd021:
The code intentionally have `?` and `&` at the end of the URL even without store settings passed, which is still a valid URL. 

This is done to avoid complexity and make the PR diff as small as possible.

Please let me know if we want to not include these characters when ssh store settings are not used.

phaer:
@aabccd021 Sorry for the delayed response. Idea sounds good to me - could you rebase?

aabccd021:
@phaer Just rebased. Could you check?

src/nixos-anywhere.sh
```diff
@@ -94,6 +95,8 @@ Options:
   use another kexec tarball to bootstrap NixOS
 * --kexec-extra-flags
   extra flags to add into the call to kexec, e.g. "--no-sync"
+* --ssh-store-settings
+  extra ssh store settings appended to the store URI e.g. "compress=true&max-connections=10"
```
phaer, line 99: I think one flag per settting, similar to how `--option` works might be a nicer interface than a single url param string?

aabccd021: Just made a change. Does it look good?

phaer: Almost, thanks for the super-quick reply!

Last nit would be that it's now a singular setting that's passed per flag, so `--ssh-store-setting` without a trailing `s` would be preferable IMO.

phaer (approved):

name-snrl:
@Mic92 I think there's an extra `shift' here, isn't there?

https://github.com/nix-community/nixos-anywhere/blob/bad98b0685cf47eaeadcaf6787da8b51cf025693/src/nixos-anywhere.sh#L287

PR #364, opened 2024-08-25
docs: cleanup

### LINKED ISSUE #362: colmena integration documentation
The [quickstart](https://github.com/nix-community/nixos-anywhere/blob/main/docs/quickstart.md) documentation mentions that the flake created by `nixos-anywhere` can be referenced by `colmena` for doing further deployments. However, there is no documentation on exactly how to import the configuration specified by the `nixos-anywhere` flake into an existing `colmena` flake. I've tried:

```
outputs.<machine-name> = { config, pkgs, ...}: {
   deployment.targetHost = "host.address";
   imports = [ inputs.nixos-anywhere-flake.nixosConfigurations.<machine-name> ];
}
```

which gives the error:
```
Could not load a value as a module, because it is of type "configuration"
```

I'm not sure exactly how to solve this problem, and in any case it would be good if the exact syntax needed to use nixos-anywhere with colmena was documented.

sedlund:
fixes #362


terraform/install/variables.tf
```diff
@@ -61,7 +61,7 @@ variable "stop_after_disko" {
 
 variable "extra_files_script" {
   type        = string
-  description = "A script file that prepares extra files to be copied to the target host during installation. The script expected to write all its files to the current directory. This directory is copied to the target host during installation to the / directory."
+  description = "A script that creates extra files for installation. The script writes all files to the current directory. During install this directory is copied to the target host's / directory"
```
phaer, line 64: I appreciate the grammar improvements, but I am not sure if this is technically correct and/or less confusing. 

I think we can't say whether the script actually *creates* the files, maybe we could say it "collects" or "gathers" those files if "prepare" feels wrong. 

Same for saying "The scripts writes all files to the current directory". That's something depending on the passed script, we just expect it to.

terraform/all-in-one.md
```diff
@@ -110,7 +110,7 @@ No resources.
...
-| <a name="input_extra_environment"></a> [extra\_environment](#input_extra_environment)                                 | Extra environment variables to be set during installation. This can be usefull to set extra variables for the extra\_files\_script or disk\_encryption\_key\_scripts                                                                                      | `map(string)`                                                          | `{}`     |    no    |
+| <a name="input_extra_environment"></a> [extra\_environment](#input_extra_environment)                                 | Extra environment variables to be set during installation. This can be useful to set extra variables for the extra\_files\_script or disk\_encryption\_key\_scripts                                                                                       | `map(string)`                                                          | `{}`     |    no    |
```
phaer, line 113: nit: trailing whitespace

README.md
```diff
@@ -93,7 +94,7 @@ suit individual requirements, see the [How To Guide](./docs/howtos/INDEX.md).
 
 **nixos-anywhere** makes use of the
 [disko](https://github.com/nix-community/disko) tool to handle the partitioning
-and formatting of the disks.
+and formatting of storage.
```
phaer, line 97: Just curious: Do you have anything specific in mind, motivating this change? I can't think of anything disko can handle which I wouldn't call "a disk" atm.

sedlund: I can drop it.  I was just thinking I haven't seen disk storage in a while, and that storage it is more generic.

terraform/all-in-one.md
sedlund: after looking i believe this is done by nix fmt to pad the table because I removed the extra 'l' in useful

sedlund:
quickstart gets a reformat to use markdown headers instead of manually using space formatting.  😇

phaer (approved):
Putting whitespace-only and content changes into different commits would make reviewing them a tiny bit easier, but looks good to me here. Thank you again for all that work on reviewing the docs.

phaer:
@mergify queue

PR #366, opened 2024-08-29
nixos-anywhere: fixes relating to `--build-on-remote`

### LINKED ISSUE #347: [bug] nix-copy should not use ssh-ng because a nix bug makes it blow up
Hi, I have encountered this nix bug https://github.com/NixOS/nix/issues/7359 when trying to use `nixos-anywhere`. I can avoid it by using `ssh://` instead of `ssh-ng://` in `nix_copy` invocations in the script. 

I think it would be a good idea to make it a default and enable it when the user asks for it.

Enzime:
- nixos-anywhere: avoid `ssh-ng://` with `nix copy`
Fixes #347 

- terraform/install: fix `build_on_remote = false`


Enzime:
I've updated the Terraform code to pass the arguments as one JSON environment variable instead :+1:

phaer (approved):
The formatter check doesn't seem happy yet. LGTM otherwise

Enzime:
Superseded by #383

PR #367, opened 2024-08-31
docs: copy updated var descriptions to other docs

sedlund:
updated some docs where the same descriptions were reused that I updated previously.


sedlund:
bots broken i guess - saw others in nix-community with same buildbot error

phaer (approved):

phaer:
> bots broken i guess - saw others in nix-community with same buildbot error

Yes, just re-triggered.

PR #369, opened 2024-09-09
allow to disable/enable phases in nixos-anywhere

Mic92:
Instead of having flags like --no-reboot or --stop-after-disko, we now can specify all phases to run.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -501,6 +501,10 @@ nixosInstall() {
   fi
 
   step Installing NixOS
+  maybeReboot=""
+  if [[ ${phases[reboot]-} == 1 ]]; then
+    maybeReboot="nohup sh -c 'sleep 6 && reboot' >/dev/null &"
+  fi
   ssh_ sh <<SSH
 set -eu ${enable_debug}
 # when running not in nixos we might miss this directory, but it's needed in the nixos chroot during installation
@@ -527,6 +531,7 @@ if command -v zpool >/dev/null && [ "\$(zpool list)" != "no pools available" ]; 
   umount -Rv /mnt/
   zpool export -a || true
 fi
+${maybeReboot}
 SSH
 
 }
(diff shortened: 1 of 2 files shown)


src/nixos-anywhere.sh
```diff
@@ -151,11 +159,33 @@ while [[ $# -gt 0 ]]; do
...
     shift
     ;;
+  --phases)
+    phases[kexec]=0
+    phases[disko]=0
+    phases[install]=0
+    phases[reboot]=0
+    IFS=, read -r -a phase_list <<<"$2"
+    for phase in "${phase_list[@]}"; do
+      if [[ ${phases[$phase]:-unset} == unset ]]; then
```
phaer, line 169: Could/should have some quotes here, i think?

src/nixos-anywhere.sh
```diff
@@ -364,132 +393,118 @@ if [[ ${is_os-n} != "Linux" ]]; then
...
-  ssh_connection="root@${ssh_host}"
-fi
-
-for path in "${!disk_encryption_keys[@]}"; do
-  step "Uploading ${disk_encryption_keys[$path]} to $path"
-  ssh_ "umask 077; cat > $path" <"${disk_encryption_keys[$path]}"
-done
-
-if [[ ${build_on_remote-n} == "y" ]]; then
-  pubkey=$(ssh-keyscan -p "$ssh_port" -t ed25519 "$ssh_host" 2>/dev/null || {
```
phaer, line 441: This (and ssh_port) where removed without replacement, as pubkey was unused anyway. Am I reading this correctly?

phaer (approved):
Code looks good to me, haven't tested this (yet) :)

src/nixos-anywhere.sh
Mic92: Yes. pubkey wasn't used anywhere.

src/nixos-anywhere.sh
Mic92: shellcheck disagrees. And the content is also not dynamic.

Mic92:
@mergify queue

PR #386, opened 2024-09-17
Re-invert behavior of --no-substitute-on-destination

chreekat:
Maybe there's a better way to do this, but it looks like removing elements from an array is kind of a pain.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -123,6 +123,7 @@ step() {
 }
 
 parseArgs() {
+  local substituteOnDestination=y
   while [[ $# -gt 0 ]]; do
     case "$1" in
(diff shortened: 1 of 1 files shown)


src/nixos-anywhere.sh
```diff
@@ -226,7 +227,7 @@ parseArgs() {
       nixOptions+=("--option" "$key" "$value")
       ;;
     --no-substitute-on-destination)
-      nixCopyOptions+=("--substitute-on-destination")
```
Mic92, line 229: I dropped the unset because it would break `--from`.

chreekat:
I don't have a spare target machine atm, but I like your version better and it "looks right" :rocket: .

Issue #389, opened 2024-09-20
Terraform: install module fails because stop_after_disko variable is not set


max-pfeiffer:
How to reproduce (main.tf):
```
module "install" {
        source            = "github.com/nix-community/nixos-anywhere//terraform/install"
        depends_on = [proxmox_vm_qemu.opentofu-test]
        flake = local.flake_file
        build_on_remote = true
        target_host       = "192.168.1.11"
        ssh_private_key = file(local.private_key)
}
```

Running `tofu apply` fails with the following error:
```
│ Error: local-exec provisioner error
│ 
│   with module.install.null_resource.nixos-remote,
│   on .terraform/modules/install/terraform/install/main.tf line 26, in resource "null_resource" "nixos-remote":
│   26:   provisioner "local-exec" {
│ 
│ Error running command '.terraform/modules/install/terraform/install/run-nixos-anywhere.sh ': exit status 1. Output: .terraform/modules/install/terraform/install/run-nixos-anywhere.sh: line 19: input[stop_after_disko]: unbound variable
│ 
╵

```

Mic92:
Fixed in https://github.com/nix-community/nixos-anywhere/pull/391

PR #390, opened 2024-09-20
Added stop_after_disko and no_reboot to script

max-pfeiffer:
This is fixing a bug when running the `run-nixos-anywhere.sh` script.

Fixes https://github.com/nix-community/nixos-anywhere/issues/389


Mic92:
I cannot find this variable in the script: https://github.com/nix-community/nixos-anywhere/blob/main/terraform/install/run-nixos-anywhere.sh

PR #398, opened 2024-09-28
nixos-anywhere: move unmount into `reboot` phase

Enzime:
This means that users can now access the mounted filesystems by omitting the `reboot` phase.

You can combine this with `--disko-mode mount` to mount and upgrade an existing system before running `nixos-enter` to run commands on it.

    $ nixos-anywhere --disko-mode mount --phases kexec,disko,install

@Mic92 I wonder if it would be worth adding a separate `unmount` phase


Mic92:
We can add one, if requested. But I don't have a good use case for this in mind just now.

Mic92:
@mergify queue

PR #407, opened 2024-10-13
feat: expose TF `var.nixos_generate_config_path`

KiaraGrouwstra:
allows invoking `--generate-hardware-config nixos-generate-config` thru TF.

usage example in an `install` / `all-in-one` module:

```
+      nixos_generate_config_path = "./hardware/hcloud-aarch64.nix";
```

... then add the file to your nixos config, e.g. by:

```
nixosConfigurations = let
  hardware_path = ./hardware/hcloud-aarch64.nix;
in {
  my_cfg = lib.nixosSystem {
    system = "x86_64-linux";
    modules = [
      (lib.optionalAttrs (lib.pathExists hardware_path) hardware_path)
    ];
  };
};
```

PR #408, opened 2024-10-13
feat: expose `--generate-hardware-config nixos-facter` thru TF

KiaraGrouwstra:
add `var.nixos_facter_path` to TF modules `install` / `all-in-one`.

c.f. https://github.com/nix-community/nixos-anywhere/pull/407, which handled `--generate-hardware-config
nixos-generate-config`.

P.S.: this PR builds upon that one, figuring it would be more elegant to use the `elif` than have two separate `if` statements.


terraform/install/variables.tf
```diff
@@ -109,3 +109,15 @@ variable "flake" {
   description = "The flake to install the system from"
   default     = ""
 }
+
+variable "nixos_generate_config_path" {
```
Mic92, line 113: We should check that the user only provides the one or to other but not both.

flake.nix
```diff
@@ -14,6 +14,13 @@
 
     # used for development
     treefmt-nix = { url = "github:numtide/treefmt-nix"; inputs.nixpkgs.follows = "nixpkgs"; };
+
+    # used in terraform to detect hardware
+    nixos-facter = {
```
Mic92, line 19: Where is actually used? nixos-facter is also in nixpkgs. I think we should prefer that to have it binary cached. Also it's actually only run on the target and we already made it part of the kexec image.

terraform/install/variables.tf
KiaraGrouwstra: hm. it [seems](https://stackoverflow.com/a/70460343/1502035) that terraform 1.9.0 can validate input variables based on another input variable's value, tho unfortunately for now that would break support for opentofu, which (up to its [latest release](https://github.com/opentofu/opentofu/releases) 1.8.3) doesn't seem to have caught up on that yet.
i'd be open to suggestions on how to best address this for now.

terraform/install/run-nixos-anywhere.sh
```diff
@@ -27,6 +27,11 @@ if [[ -n ${input[flake]} ]]; then
 else
   args+=("--store-paths" "${input[nixos_partitioner]}" "${input[nixos_system]}")
 fi
+if [[ -n ${input[nixos_generate_config_path]} ]]; then
+  args+=("--generate-hardware-config" "nixos-generate-config" "${input[nixos_generate_config_path]}")
```
Mic92, line 31: Here we should check that nixos_facter_path is not set.

KiaraGrouwstra: thanks, i added the check now!

terraform/install/variables.tf
```diff
@@ -109,3 +109,15 @@ variable "flake" {
...
+  type        = string
+  description = "Path to which to write a `hardware-configuration.nix` generated by `nixos-generate-config`."
```
Mic92, line 115:
```suggestion
  description = "Path to which to write a `hardware-configuration.nix` generated by `nixos-generate-config`. This option cannot be set at the same time as `nixos_facter_path`"
```

terraform/install/variables.tf
```diff
@@ -109,3 +109,15 @@ variable "flake" {
...
+  default     = ""
+}
+
+variable "nixos_facter_path" {
+  type        = string
+  description = "Path to which to write a `facter.json` generated by `nixos-facter`."
```
Mic92, line 121:
```suggestion
  description = "Path to which to write a `facter.json` generated by `nixos-facter`. This option cannot be set at the same time as `nixos_generate_config_path`."
```

Mic92:
We don't have a integration test for this. @KiaraGrouwstra could you test this one more time to make sure nothing is broken. Otherwise I would need to run a test later.

KiaraGrouwstra:
@Mic92

> We don't have a integration test for terraform. @KiaraGrouwstra could you test this one more time to make sure nothing is broken. Otherwise I would need to run a test later.

tho in retrospect separate from TF, i pushed an update to the nixos-anywhere bash script to ensure any directories of hardware configuration files would be created, if needed.

that made it a bit more robust, and had it go thru for me. as in #407, in my use i am still checking if `pathExists` again using `(lib.optionalAttrs (lib.pathExists reportPath) { config.facter = { inherit reportPath; }; })`. that is, if using flakes + git, the user is currently still expected to manually stage the generated file to git for the nixos config to actually pick it up.
(that said, that particular weakness too i think seems not specific to TF.)

Mic92:
The git add we already do: https://github.com/nix-community/nixos-anywhere/blob/3fb834da466b75300a9141f9a68869c56fbd28f3/src/nixos-anywhere.sh#L458

Mic92:
@mergify queue

Issue #409, opened 2024-10-15
feat: How to connect with different user after reboot?


gabyx:
I have VM that has a user `ubuntu`.

 I can run `nixos-anywhere --flake ... 11.11.11.11`  (which actually works on engine.switch.ch which is really great). The `~/.ssh/config` is:

```ini
Host 11.11.11.11
    IdentityFile ~/.ssh/vm
    IdentitiesOnly yes
    User ubuntu
```

However once it reboots -> I can no longer connect because the user changed to `root` =). It works if I in the meantime change the `User ubuntu` to `User root`.

Is there a way to somehow trick `nixos-anywhere` to use `root` after install or some other ssh options?

gabyx:
Or even maybe `nixos-anywhere` could provide an option: `--post-kexec-ssh-user` or even `--post-kexec-ssh-hostname` where the later gives the opportunity to use `--post-kexec-ssh-hostname afterinstall` with an `~/.ssh/config` as;

```ini
Host 11.11.11.11
    IdentityFile ~/.ssh/vm
    IdentitiesOnly yes
    User ubuntu

Host afterinstall
    IdentityFile ~/.ssh/vm
    IdentitiesOnly yes
    User root
```

sedlund:
It should kexec onto the new kernel (this is not a reboot) and nixos-anywhere will reconnect to the machine using generated ssh keys from prior to kexec and install nixos.  is this not happening?

your nixosConfiguration flake should have configured the user and ssh key (if that is your method to connect) to use after reboot.

by after reboot am I correct to think that NixOS is installed at this point?  If that is the case nixos-anywheres job is [complete](https://github.com/nix-community/nixos-anywhere/blob/main/docs/quickstart.md#finished).

gabyx:
> your nixosConfiguration flake should have configured the user and ssh key (if that is your method to connect) to use after reboot.
Jeah: My ssh config is the same as  https://github.com/nix-community/nixos-anywhere-examples/blob/main/configuration.nix#L26

First I do `nixos-anywhere --flake ubuntu@host` and then the installation after `kexec` fails to reconnect, I can leave it trying to reconnect -> and change `User` to `root` in the `.ssh/config` and then it suddenly works and I see all nix store paths getting copied and the install finished successfully =)

sedlund:
please provide the full logs with `--debug` enabled along with the command you are running.

it should be trying to reconnect with the user provided to install as and using the generated key that it installed.

[...]

PR #411, opened 2024-10-18
docs: add hint to not use `IdentitiesOnly` in SSH config

gabyx:
- Issue #409 gave rise to adding some hints on the SSH config to not run into the same troubles.

Thanks @sedlund!

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -66,6 +66,18 @@ line that reads:
 Replace the text `CHANGE` with your own SSH key. This is crucial, as you will
 not be able to log into the target machine post-installation without it.
 
+**Note:** If you use an `~/.ssh/config` entry **make sure the entry for your machine **does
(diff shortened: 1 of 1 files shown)


phaer:
Formatting seems to be off, see treefmt but also the stray `\*\*`. 

Otherwise LGTM. (Might help to at least optionally use a local ssh agent instead of the generated key, but that's for another PR :))

docs/quickstart.md
```diff
@@ -66,6 +66,18 @@ line that reads:
 Replace the text `CHANGE` with your own SSH key. This is crucial, as you will
 not be able to log into the target machine post-installation without it.
 
+**Note:** If you use an `~/.ssh/config` entry **make sure the entry for your machine **does
+not include `IdentitiesOnly yes`\*\* since that will block the on-the-fly generated
```
gabyx, line 70:
```suggestion
not include `IdentitiesOnly yes` since that will block the on-the-fly generated
```

PR #414, opened 2024-10-23
terraform: add `var.extra_build_env_vars`, closes #413

### LINKED ISSUE #413: exposing terraform info back to nixos configurations
in order to prevent duplication and hard-coding, it would be nice if we had a way to pass info from TF to the nixos configurations.

before nixos-anywhere i used [teraflops](https://github.com/aanderse/teraflops), which exposed a static TF `show -json` dump thru a `specialArgs` parameter [`resources`](https://github.com/aanderse/teraflops#special-arguments).
now, they can use such a static dump as they deploy the nix bits after terraform. as nixos-anywhere deploys nixos during terraform execution, here such info would instead need to be passed from within the terraform.

this could be facilitated in different ways, e.g. opinionated like teraflops (exposing such info in a given format), versus more generically by just passing thru more nix arguments, while leaving the details to the user.

there would also be considerations there between exposing info in a pure vs impure way:

- passing info in a pure way, as files
  - pros:
    - may be preferred by nix
  - cons:
    - irrelevant changes in ephemeral TF info may show up in git (if not also triggering unnecessary rebuilds)
- passing info in an impure way, as env vars
  - pros:
    - no fighting flakes about needing to stage files to git
  - cons:
    - if this is impure, i'm not sure if nix would still know whether it should rebuild if just the contents change

KiaraGrouwstra:
This PR adds a Terraform input variable I so far named `extra_build_env_vars` (suggestions welcome).
This allows passing in a map of attributes from Terraform to expose to nixos-anywhere's build step to make them available as environment variables within a NixOS configuration.

For this implementation I opted to leave actually passing the info to the user, so people stay in control over what info gets exposed where.
I opted to focus on allowing to pass info thru environment variables here figuring that would be less prone to fighting flakes over what is staged to Git, as committing info to be passed here may be undesirable both for ephemeral noise (if not also over sensitive info - tho any info passed this way lands in `/nix/store/` as well).

To pull this off I used `--impure` plus `eval` - feedback on this would be great.

On the Terraform side as a consumer, to best enjoy this I have had to resort to some more involved logic as well, that gets easy when using Terraform from Nix. As an example (based on Hetzner Cloud), using the [`lib.tfRef` function](https://terranix.org/documentation/terranix-vs-hcl.html#escaping-in-single-line-strings) from Terranix I would use something like:

```nix
let
  servers = ...;
  variable = ...;
  data = ...;
  resource = ...;
in
{
  inherit variable data resource;
  module =
    lib.mapAttrs (server_name: _server_config: let
    in {
      # pin module version by nix flake inputs
      source = "github.com/numtide/nixos-anywhere?ref=${inputs.nixos-anywhere.sourceInfo.rev}/terraform/all-in-one";
      ...
      extra_build_env_vars = {
        # all variables
        # TF_VARS = lib.strings.toJSON (lib.mapAttrs (k: _: lib.tfRef "jsonencode(var.${k})") variable);
        # non-sensitive variables
        TF_VARS = lib.tfRef "jsonencode(${lib.strings.toJSON (lib.mapAttrs (k: _: lib.tfRef "var.${k}") (lib.filterAttrs (_k: v: !(v ? sensitive && v.sensitive)) variable))})";
        TF_DATA = lib.tfRef "jsonencode(${lib.strings.toJSON (lib.mapAttrs (type: instances: lib.mapAttrs (k: _: lib.tfRef "data.${type}.${k}") instances) data)})";
        TF_RESOURCES = lib.tfRef "jsonencode(${lib.strings.toJSON (lib.mapAttrs (type: instances: lib.mapAttrs (k: _: lib.tfRef "resource.${type}.${k}") instances) resource)})";
        TF_SERVER = lib.tfRef "jsonencode(resource.hcloud_server.${server_name})";
        SERVER_NAME = server_name;
      };
    })
    servers;
}
```

You can then verify their contents from your `nixosConfigurations` like:

```nix
          lib.nixosSystem {
            inherit system;
            modules = [
              {
                environment.etc = {
                  TF_VARS.text = builtins.getEnv "TF_VARS";
                  TF_DATA.text = builtins.getEnv "TF_DATA";
                  TF_RESOURCES.text = builtins.getEnv "TF_RESOURCES";
                  TF_SERVER.text = builtins.getEnv "TF_SERVER";
                  SERVER_NAME.text = builtins.getEnv "SERVER_NAME";
                };
              }
            ];
          };
```

-> `ls /etc/` -> `cat /etc/TF_SERVER`

(in practice one would instead probably parse these back to nix values to grab specific properties from, e.g. like: `lib.strings.fromJSON (builtins.getEnv "TF_SERVER")`)


Mic92:
@mergify queue

terraform/nix-build/nix-build.sh
```diff
@@ -1,15 +1,16 @@
...
-declare file attribute nix_options
-eval "$(jq -r '@sh "attribute=\(.attribute) file=\(.file) nix_options=\(.nix_options)"')"
+declare file attribute nix_options environment
+eval "$(jq -r '@sh "attribute=\(.attribute) file=\(.file) nix_options=\(.nix_options) environment=\(.environment)"')"
 options=$(echo "${nix_options}" | jq -r '.options | to_entries | map("--option \(.key) \(.value)") | join(" ")')
+vars=$(echo "${environment}" | jq -r "to_entries | map(\"\(.key)='\(.value)'\") | join(\" \")")
 if [[ -n ${file-} ]] && [[ -e ${file-} ]]; then
   # shellcheck disable=SC2086
-  out=$(nix build --no-link --json $options -f "$file" "$attribute")
+  out=$(eval "env ${vars} nix build --no-link --json --impure $options -f '$file' '$attribute'")
```
Mic92, line 10: Mhm this is not so nice.

Have you considered the following approach:

https://github.com/NixOS/nixos-wiki-infra/blob/main/terraform/nixos-wiki/nixos_vars.tf

Basically one can use terraform to generate a json file and using `git add` to make sure it's recognized by the flake. Maybe we can encode this pattern into the terraform module instead of using `builtins.getEnv`?
The big upside on this is that, you are no longer forced to always have to use terraform to build or deploy your nixos configuration.

Mic92: Just imagine you want your machines to be build by CI. Before you could just do `nixos-rebuild build --flake .mymachine`, now you have to do some sort of `terraform plan`.

terraform/nix-build/variables.tf
```diff
@@ -14,3 +14,9 @@ variable "nix_options" {
   description = "the options of nix"
   default = {}
 }
+
+variable "extra_build_env_vars" {
```
Mic92, line 18: I think in general it would be fine to have environment variables or flags passed to nixos-rebuild. This could be used for injecting binary cache configuration.

terraform/nix-build/nix-build.sh
Mic92: Interface wise I would propose a `nix-file` module that would use `git add --intent-to-add --force -- ./file >/dev/null 2>&1 || true` to make sure that it is added to your git repository (if one exists). It would take `content` as a parameter:

```
module "nixos_vars" { 
   source  = "../nix-file"
   content = jsonencode({})
   filename = "${path.module}/nixos-vars.json"
}
```

KiaraGrouwstra: hm. if one could make a file without terraform, why couldn't one similarly make env vars, if the nixos configuration is indeed made to require them?

locally such a `nix-file` module might give some noise in version control, tho one could opt to just never use `git add .` anymore. whereas deploying from CI might circumvent that, it would seem a bit annoying in a development setting. going by file may have use-cases as well, tho it'd be nice to see a route more amenable to the common use-case of git + flakes + development supported as well.

on `--impure`, maybe a conditional could be added.

terraform/nix-build/variables.tf
KiaraGrouwstra: thanks, i'll check it out. i still got a few question marks around say how that would handle conditionals based on environment variables, but i guess i can just try that.

terraform/nix-build/nix-build.sh
Mic92: As we discussed it's possible to achieve your goal without impure evaluation. Check how to import json into a nixos configuration: https://github.com/nix-community/disko/blob/09a776702b004fdf9c41a024e1299d575ee18a7d/install-cli.nix#L56

disko-install still adds `--impure`. However it's possible to use getFlake also in a pure evaluation: Check out this code in clan: https://git.clan.lol/clan/clan-core/src/commit/bcf2cd1814bbef57c74b0b4eec8f6a7cc0479475/pkgs/clan-cli/clan_cli/machines/machines.py#L229

Mic92: https://github.com/KiaraGrouwstra/nixos-anywhere/commit/75e09d1681d8afb129c2b9f385859c84a5e77a27 also looks good.

KiaraGrouwstra: @Mic92 thanks for elaborating - i originally hadn't quite figured out the tempfile thing. i'll try and see if I can use the `readFile` you mentioned approach to resolve the flake+git friction.

Mic92:
This is obsolete now.

PR #417, opened 2024-10-31
terraform/nix-build.sh: prevent error on empty nix_options

KiaraGrouwstra:
currently, when `var.nix_options` is not set, the jq map action yields `null`, which ends up missing up. using a conditional here, we may instead account for this case such as to set it to empty string and avoid the issue.


Mic92:
@mergify queue

PR #418, opened 2024-10-31
terraform: add `var.special_args`, closes #413, supersedes #414.

### LINKED ISSUE #413: exposing terraform info back to nixos configurations

KiaraGrouwstra:
This PR adds a Terraform input variable named `special_args`.

This allows passing in a map from Terraform to expose to NixOS's `specialArgs` at build-time.
Example usage, in this case presuming deployment to a Hetzner Cloud server (`resource.hcloud_server`):

```nix
let
  servers = ...;
  variable = ...;
  data = ...;
  resource = ...;
in
{
  inherit variable data resource;
  module =
    lib.mapAttrs (server_name: _server_config: let
    in {
      # pin module version by nix flake inputs
      source =
"github.com/numtide/nixos-anywhere?ref=${inputs.nixos-anywhere.sourceInfo.rev}/terraform/all-in-one";
      ...
      special_args = {
        tf = {
          inherit server_name;
          # all variables
          # var = lib.mapAttrs (k: _: lib.tfRef "var.${k}") variable;
          # non-sensitive variables
          var = lib.mapAttrs (k: _: lib.tfRef "var.${k}")
  (lib.filterAttrs (_k: v: !(v ? sensitive && v.sensitive)) variable);
          data = lib.mapAttrs (type: instances: lib.mapAttrs (k: _:
  tfRef "data.${type}.${k}") instances) data;
          resource = lib.mapAttrs (type: instances: lib.mapAttrs (k:
  _: tfRef "resource.${type}.${k}") instances) resource;
          server = lib.tfRef "resource.hcloud_server.${server_name}";
        };
      };
    })
    servers;
}
```

You can then use these in your `nixosConfigurations`, in this example thru the `tf` argument.

As a note on security, information passed this way _will_ hit `/nix/store/`. As such, the above usage example has defaulted to omitting TF variables marked as sensitive.

This PR incorporates ideas from:

- @aanderse, who implemented a similar feature in [teraflops](https://github.com/aanderse/teraflops) that inspired this PR.
- @Mic92, who suggested (see https://github.com/nix-community/nixos-anywhere/pull/414) to extend the original `lib.nixosSystem` call to pass in info without `--impure` or staging to Git.
- @getchoo, who suggested getting the NAR hash by `nix flake prefetch` over `getFlake`, fixing an 'unlocked flake reference' error on (non-Lix) Nix.
- @threddast, who suggested to use TF's `any` type to automate serializing.

An [alternative design](https://github.com/nix-community/nixos-anywhere/compare/main...KiaraGrouwstra:nixos-anywhere:tf-info-to-wrapper#diff-2e2429dde4812f0b50c784e8d4c8b93cc9faeb52cce4747733200f65ea5c2bbb) suggested by @Mic92 involved passing the information not directly, but rather thru a file. The idea would be that this might help reduce the risk of stack overflows, tho I have imagined (perhaps naively) that TF info has tended not to get too big, whereas I also had a bit more trouble getting that approach to work properly so far (involving both NARs that would suddenly mismatch again, while I'd also yet to test if one could put such files in `.gitignore`).


Mic92:
Oh wait, this is computing the nar from the flake that defines the NixOS config and than passes in the special args as a command line argument again. The idea was to do the same nar calculation for the JSON file that is passed in as an input and than just use getFlake on the original configuration and pass in the JSON file path via readFile. This than avoids large string I can follow up later with more specific, currently on my phone.

KiaraGrouwstra:
@Mic92 yes, that sounds like this earlier iteration where i tried `specialArgs` thru a file:

> An [alternative design](https://github.com/nix-community/nixos-anywhere/compare/main...KiaraGrouwstra:nixos-anywhere:tf-info-to-wrapper#diff-2e2429dde4812f0b50c784e8d4c8b93cc9faeb52cce4747733200f65ea5c2bbb) suggested by @Mic92 involved passing the information not directly, but rather thru a file. The idea would be that this might help reduce the risk of stack overflows, tho I have imagined (perhaps naively) that TF info has tended not to get too big, whereas I also had a bit more trouble getting that approach to work properly so far (involving both NARs that would suddenly mismatch again, while I'd also yet to test if one could put such files in `.gitignore`).

KiaraGrouwstra:
i did consider a hybrid, where an optional filename argument could be set such as to go thru that... but felt like the feature as-is seemed functional enough to submit already

Mic92:
Okay. I will try out later.

KiaraGrouwstra:
extra notes:

- this is only supported using flakes currently
- it would also be possible to split this up for `toplevel` vs `diskoScriptNoDeps` - so far i had not done that

threddast:
I'm trying to test this, but I'm always getting this error

```
nix build --expr "builtins.getFlake ''git+file://${flake_dir}?narHash=sha256-0000000000000000000000000000000000000000000=''"     
error:
       … while calling the 'getFlake' builtin
         at «string»:1:1:
            1| builtins.getFlake ''git+file:///home/threddast/devel/cloud-infra?narHash=sha256-0000000000000000000000000000000000000000000=''
             | ^

       error: cannot call 'getFlake' on unlocked flake reference 'git+file:///home/threddast/devel/cloud-infra?narHash=sha256-0000000000000000000000000000000000000000000=', at «none»:0 (use --impure to override)

```
Even though I have a flake.lock committed in git.
I get the same also when trying with the flake from https://codeberg.org/kiara/e2ed-hetzner
My Nix version is 2.24.10

KiaraGrouwstra:
@threddast thank you for your feedback.
i did run into issues like that as well using Nix, tho using Lix seemed to address this.
i guess it might be possible to bisect Nix versions on that to see if/where it gets fixed.
a [search in the Nix repo](https://github.com/search?q=repo%3ANixOS%2Fnix+cannot+call+%27getFlake%27+on+unlocked+flake+reference&type=issues) shows similar symptoms had been reported previously, tho it seems the present bug (regression?) seems to pop up even with the NAR hash specified and all.
i just kinda presume nix would fix it eventually, whereas lix seems just more stable already.

KiaraGrouwstra:
@threddast courtesy of @getchoo's advice on lix matrix, i pushed an update to calculate the NAR hash differently.
it's looking like that should address the issues experienced when running this with Nix.

terraform/nix-build/main.tf
```diff
@@ -9,6 +9,7 @@ data "external" "nix-build" {
     attribute = var.attribute
     file = var.file
     nix_options = local.nix_options
+    special_args = var.special_args
```
threddast, line 12: What about adding 
locals {
  nix_options = jsonencode({
    options = { for k, v in var.nix_options : k => v }
  })
  special_args = jsonencode({
    tf = { for k, v in var.special_args : k => v }
  })
}

threddast (commented):
Thank you, I tested the module and it works really well now.
Just a suggestion, would it make sense to move the `jsonencode` inside the module, similar to the `nix_options` variable?

terraform/nix-build/main.tf
KiaraGrouwstra: i agree not needing to manually `jsonencode` would seem nicer, and had tried it as well.
unfortunately, while it works for this dummy scenario, TF seems to offer no JSON-like type for input variables. it allows your example because it seems 'map of strings to strings' as a valid type, but i had not managed to get it to work for deeper structures than that.
TF's resource/data info then wouldn't fit in there without manually serializing every value. at that point, i felt like just manually serializing the whole thing once would seem the lesser evil. :(

KiaraGrouwstra: Thank you, looks like that worked! Pretty happy about this change! :see_no_evil: :v:

terraform/all-in-one.md
```diff
@@ -33,6 +33,17 @@ module "deploy" {
...
   }]
+  # Optional, arguments based to special_args here will be available from a NixOS module via the terraform argument:
+  # { terraform, ... }: {
+  #    networking.interfaces.enp0s3.ipv4.addresses = [{ address = terraform.ip;  prefixLenght = 24; }];
+  # }
+  # Note that this will means that your NixOS configuration will always depend on terraform!
+  # Skip to `Pass data persistently to the NixOS` for an alternative approach
+  #special_args = {
+  #  terraform = {
+  #    ip = "192.0.2.0"
```
Mic92, line 44: @KiaraGrouwstra is this example correct? I derived from your pull request message.

Mic92:
If you could also solve the merge conflict, than we could finally merge this.
Thank you!

terraform/all-in-one.md
KiaraGrouwstra: nit: the user can pass arguments under any name (or multiple even), rather than just `terraform` - otherwise this sounds right!

Mic92:
@mergify queue

PR #422, opened 2024-11-08
fix(disk-encryption-keys): ensure the target dir exists

dudeofawesome:
When using `--disk-encryption-keys`, copying will silently fail if the target file location's parent directory does not exist. 

This PR adds a `mkdir -p` to ensure that the target dir does exist.


Mic92:
@mergify queue

PR #424, opened 2024-11-12
hardware config: fix git add logic

picnoir:
Fixing the git add hardware config logic in case the user provides a relative path. We're doing two things:

1. Failing loudly if the git add command fails. At that point, we already know git is in $PATH. If the git add fails, the evaluation will likely fail in a flake setting. Better fail early with a helpful git error message.
2. We're resolving the hardware config path in the CLI parsing layer to make sure this path has the meaning the user intented it to have.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -176,7 +176,7 @@ parseArgs() {
         abort "Unknown hardware config backend: $2"
         ;;
       esac
-      hardwareConfigPath=$3
+      hardwareConfigPath="$(realpath "$3")"
       shift
       shift
       ;;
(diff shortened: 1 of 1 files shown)


picnoir:
The CI failure is related to a test case where we have git but the local repo is not git-tracked.

I guess we should check if the local repo is git-tracked before failing?

picnoir:
CI fixed. This is ready for review.

src/nixos-anywhere.sh
```diff
@@ -455,9 +455,9 @@ generateHardwareConfig() {
 
   # to make sure nix knows about the new file
   if command -v git >/dev/null; then
-    pushd "$(dirname "$hardwareConfigPath")"
-    git add --intent-to-add --force -- "$hardwareConfigPath" >/dev/null 2>&1 || true
-    popd
+    if git rev-parse --is-inside-work-tree >/dev/null; then
```
Mic92, line 458: So does this mean users have to change into the repository for this to work.
I can imagine that users would do something like
`nixos-anywhere --flake $HOME/git/my-config`

picnoir: Riiight, I get the popd part now!

picnoir: Fixed

Mic92:
@mergify queue

PR #427, opened 2024-11-21
add var.ssh_options to pass additional SSH options to nixos-rebuild

threddast:
This PR adds a Terraform input variable named `ssh_options` to nixos-rebuild (and to the rebuild part of nixos-anywhere)

The problem I'm trying to solve is being able to use `gcloud compute ssh` to rebuild NixOS. `gcloud compute ssh` is a wrapper around `ssh` that sets a bunch of ssh options to connect to a GCP VM, even if there is no direct connection to it.

Example usage:
```
module "deploy" {
  source       = "github.com/nix-community/nixos-anywhere//terraform/nixos-rebuild"
  nixos_system = module.system-build.result.out
  target_host = "target"
  ssh_options = {
    IdentityFile          = "/home/user/.ssh/google_compute_engine"
    CheckHostIP           = "no"
    HashKnownHosts        = "no"
    HostKeyAlias          = "compute.123456789"
    IdentitiesOnly        = "yes"
    StrictHostKeyChecking = "yes"
    UserKnownHostsFile    = "/home/user/.ssh/google_compute_known_hosts"
    ProxyCommand          = "/nix/store/nrzkzg0if8p9ak18070x8mj6clbcvdm7-python3-3.11.9-env/bin/python -S /nix/store/lann7mq6gpyl7q3d3hq6d93jr69pgm0v-google-cloud-sdk-475.0.0/google-cloud-sdk/lib/gcloud.py compute start-iap-tunnel test-vm %p --listen-on-stdin --project=my-project --zone=europe-west4-a --verbosity=warning"
    ProxyUseFdpass        = "no"
  }
}
```

I'm using a file to pass the SSH options because of this bug in Nix: https://github.com/NixOS/nix/issues/5181. The content of NIX_SSHOPTS is passed to [this tokenizer](https://github.com/NixOS/nix/blob/fa17927d9d75b6feec38a3fbc8b6e34e17c71b52/src/libutil/strings-inline.hh#L23) which splits by spaces and doesn't take quoted substrings into account.
This means that something like `NIX_SSHOPTS='-o ProxyCommand="my-ssh-command foo bar"'` wont't work unless the tokenizer is changed upstream. 
The disadvantage of passing options as a file is that it will break the terraform module for people relying on their `.ssh/config`, unless they copy the ssh options from `.ssh/config` to the terraform module. I understand this might be too much of a downside and we might want to change the tokenizer instead.


terraform/nixos-rebuild/deploy.sh
```diff
@@ -14,14 +14,16 @@ TARGET_PORT=$4
...
-
 workDir=$(mktemp -d)
 trap 'rm -rf "$workDir"' EXIT
 
-sshOpts=(-p "${TARGET_PORT}")
-sshOpts+=(-o UserKnownHostsFile=/dev/null)
-sshOpts+=(-o StrictHostKeyChecking=no)
+sshConfigFile="$workDir/ssh_config"
+cat >"$sshConfigFile" <<EOF
+Host $TARGET_HOST
```
Mic92, line 22: Yes. As you have rightfully said, this will be an issue for users configuring private ssh keys and jump hosts in their ssh_config.

Mic92: Here is a workaround for you specifically, could you not pass in -F in your case instead? terraform also allows to write files, no?

threddast: > Here is a workaround for you specifically, could you not pass in -F in your case instead? terraform also allows to write files, no?

Yes this would work too. I changed the PR so that additional options are passed as a string to `NIX_SSHOPTS`

> Maybe we can add support for passing in an ssh config file?
I think passing flags is more flexible. As you suggested we can pass `-F` if a config file is needed

terraform/nixos-rebuild/deploy.sh
```diff
@@ -20,8 +20,7 @@ workDir=$(mktemp -d)
 trap 'rm -rf "$workDir"' EXIT
 
 sshOpts=(-p "${TARGET_PORT}")
-sshOpts+=(-o UserKnownHostsFile=/dev/null)
-sshOpts+=(-o StrictHostKeyChecking=no)
```
Mic92, line 24: This changes the current behavior in nixos-anywhere, if you are not using terraform.

Mic92:
This one will be also useful: https://github.com/NixOS/nix/pull/12020

eliecharra:
I'm also interested by this change since it can also cover another usecase 👇🏻 

I'm using passwordless sudo with `security.pam.sshAgentAuth.enable = true;` in order to avoid having NOPASSWD directives in my sudoers. This require me to add the `-A` flag to ssh to forward my agent.
I can use a workaround by setting a ssh config like below, but It would be simpler to just pass an additional `-A` flag there 😃 

```
 Host target_host
   ForwardAgent yes
```

threddast:
Closing this as I'm not using `gcloud compute ssh` with `nixos-anywhere` anymore.

PR #430, opened 2024-11-29
nixos-anywhere: allow users to specify custom path NixOS configurations

Enzime:
This is useful for users who use a similar `extendModules` pattern like `virtualisation.vmVariant`, where an extended NixOS configuration lives inside an existing configuration:

    nixos-anywhere --flake .#nixosConfigurations.alpha.config.virtualisation.vmVariant

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -313,6 +313,11 @@ parseArgs() {
       echo 'For example, to use the output nixosConfigurations.foo from the flake.nix, append "#foo" to the flake-uri.' >&2
       exit 1
     fi
+
+    # Support .#foo shorthand
+    if [[ $flakeAttr != nixosConfigurations.* ]]; then
+      flakeAttr="nixosConfigurations.\"$flakeAttr\".config"
+    fi
   fi
 
 }
@@ -363,7 +368,7 @@ runVmTest() {
     --no-link \
     -L \
     "${nixOptions[@]}" \
-    "${flake}#nixosConfigurations.\"${flakeAttr}\".config.system.build.installTest"
+    "${flake}#${flakeAttr}.system.build.installTest"
 }
 
 uploadSshKey() {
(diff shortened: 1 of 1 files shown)


src/nixos-anywhere.sh
```diff
@@ -313,6 +313,19 @@ parseArgs() {
...
+
+    # We check if nixosConfigurations."<flakeAttr>".config.system.build.toplevel exists
+    # to allow users to use the shorthand $flake#foo. We use `nix flake metadata $flake`
+    # to add $flake to the store, otherwise relative paths will fail when used with
+    # `builtins.getFlake`. We use `nix eval --expr ... ? toplevel` to avoid instantiating
+    # the derivation as this may require building the entire system if the user uses IFD
+    # like through `system.replaceDependencies`.
+    if [[ $(nix eval \
+      --expr \
+      "(builtins.getFlake \"$(nix flake metadata "$flake" --json | jq -r '.path')\").nixosConfigurations.\"$flakeAttr\".config.system.build ? toplevel" \
```
Mic92, line 325: The issue is that even this might already trigger FOD and fail (sometimes people have weird flake setups). So here my proposal. We check for `flakeAttr` for the `nixosConfiguration` prefix and if it's not present, we add it. Otherwise we use the raw value and just append `config.system.build.toplevel`.

Enzime: Yeah that sounds good 👍

Enzime: How does this look now?

src/default.nix
```diff
@@ -29,6 +30,7 @@ let
     gnused # needed by ssh-copy-id
     sshpass # used to provide password for ssh-copy-id
     gnutar # used to upload extra-files
+    jq # used for checking if flakeAttr exists
```
Mic92, line 33:
```suggestion
```

src/default.nix
```diff
@@ -11,6 +11,7 @@
 , findutils
 , gnused
 , sshpass
+, jq
```
Mic92, line 14:
```suggestion
```

Enzime:
@mergify queue

PR #432, opened 2024-12-05
nixos-anywhere: add --target-host option

zimbatm:
Mirror the option from nixos-rebuild, so both have the same signature:

* Bootstrap: `nixos-anywhere --flake .#targethost --target-host root@1.2.3.4`

* Deploy: `nixos-rebuild --flake .#targethost --target-host root@1.2.3.4 switch`

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,12 +1,14 @@
 # CLI
 
 ```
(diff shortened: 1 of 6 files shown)


Mic92:
@mergify queue

Issue #438, opened 2024-12-17
The generated hardware-configuration file has CRLF terminators


plietar:
If I run `nixos-anywhere` with `--generate-hardware-config nixos-generate-config hardware-configuration.nix`, the produced file has CRLF terminators, which I would not normally expect.

```
$ nix run github:nix-community/nixos-anywhere -- --flake '.#<hostname>' root@<hostname> --generate-hardware-config nixos-generate-config hardware-configuration.nix --phases "kexec"
$ file hardware-configuration.nix
hardware-configuration.nix: Unicode text, UTF-8 text, with CRLF line terminators
```

While it is certainly not a big issue (Nix doesn't care about line termination) and I can easily work around it, it is surprising and causes large diffs if I compare the file to a previous version I had generated not with nixos-anywhere.

I've tracked the issue down to the usage of `-t` in the ssh options:
```
$ ssh root@<hostname> nixos-generate-config --show-hardware-config --no-filesystems | file -
/dev/stdin: Unicode text, UTF-8 text
$ ssh -t root@<hostname> nixos-generate-config --show-hardware-config --no-filesystems | file -
/dev/stdin: Unicode text, UTF-8 text, with CRLF line terminators
```

sedlund:
This should be fixed on main branch.  If you would like to test, please do so and close this issue accordingly.

plietar:
I've finally gotten around to trying this. Seems all good now. Thanks!

PR #439, opened 2024-12-18
fix(nixos-generate-config): dont output carriage returns in output files

### LINKED ISSUE #423: nixos-facter not available
I am trying to install to a machine that is running the minimal nixos iso and generate the hardware config using nixos-facter. However, I keep getting
```
aborted: nixos-facter is not available in booted installer. You may want to boot an installer image from here instead: https://github.com/nix-community/nixos-images
```
I have tried visiting the page but couldn't find anything on there that helped.

sedlund:
`ssh -t` is being used by `runSsh` which causes new lines in output designed for a terminal.  It could also output escape sequences.

https://unix.stackexchange.com/a/420438

this change fixes this problem (#438) by only outputting new lines.  but it might be worth removing ssh password entry as an option and the tty allocation, and force the use of sshpass for passwords.

There is a FIXME note regarding the output of the `nixos-facter` with _strange_ characters.  This may resolve that as well. 

Also updated the message about `nixos-facter` not being found to hopefully clarify and resolve #423 around the `--generate-hardware-config` option.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -72,7 +72,7 @@ Options:
   Also supports variants:
   nixos-anywhere --flake .#nixosConfigurations.mymachine.config.virtualisation.vmVariant
 * --target-host <ssh-host>
-  specified the SSH target host to deploy onto.
+  set the SSH target host to deploy onto.
 * -i <identity_file>
   selects which SSH private key file to use.
 * -p, --ssh-port <ssh_port>
@@ -332,6 +332,9 @@ parseArgs() {
 }
 
 # ssh wrapper
+runSshNoTty() {
+  ssh -i "$sshKeyDir"/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no "${sshArgs[@]}" "$sshConnection" "$@"
+}
 runSshTimeout() {
   timeout 10 ssh -i "$sshKeyDir"/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no "${sshArgs[@]}" "$sshConnection" "$@"
 }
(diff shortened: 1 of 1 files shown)


Mic92:
What version did you use for nixos-facter. We switched logging libraries at some point, which resolved an issue with ANSI escapes.

sedlund:
should have quoted:

https://github.com/nix-community/nixos-anywhere/blob/bf4c0c50d65b4f1cf13972da7feaa4108879abc9/src/nixos-anywhere.sh#L457-L459

I was unsure what 'weird characters' referred to.

So maybe what you refer to fixed that, but CRLF in output persists with tty allocation with ssh.

I tested with:

```sh
ssh -t localhost nixos-generate-config --show-hardware-config --no-filesystems > ngc-old.out
ssh -t localhost "stty nl; nixos-generate-config --show-hardware-config --no-filesystems" > ngc-new.out
ssh -t localhost sudo nixos-facter > facter-old.out
ssh -t localhost "stty nl; sudo nixos-facter" > facter-new.out

fd -e out -x nix shell nixpkgs#tinyxxd -c xxd {} {.}.xxd
```

```console
file *.out
facter-new.out: JSON text data
facter-old.out: JSON text data
ngc-new.out:    Unicode text, UTF-8 text
ngc-old.out:    Unicode text, UTF-8 text, with CRLF line terminators
```

```console
delta --side-by-side ngc-old.xxd ngc-new.xxd
delta --side-by-side facter-old.xxd facter-new.xxd
```

this is `qpqn0b91qbh879snbqpr04z8rdi7zkq8-nixos-facter-0.3.0`.

diffs show CRLF in the `old` output

Mic92:
We can remove the comment. It was caused by charm bracelet's logging library: https://github.com/numtide/nixos-facter/pull/119
It wrote some ansi escape at the very beginning, which broke the json output. However this was removed.

Mic92:
Are you running on macOS on the host by chance?

Mic92:
But I think you are write, we shouldn't use `-t` for these commands. It's only needed for the installation because we need support for sudo and other password prompts.

Mic92:
Could you try to remove it and see if this is fixing the issue?

sedlund:
> Are you running on macOS on the host by chance?

my client is NixOS-wsl.  dont know what reporter in #438 is running.

created a new runSshNoTty function that omits the tty request.

tested outputs of both `nixos-facter` and `nixos-generate-config` on a hetzner vm - no CRLF in this patched version.

Mic92:
@Enzime this is also what we saw the other day.

Mic92:
@mergify queue

Issue #283, opened 2024-02-20
The `--extra-files <file...>` command is unintuitive


Qubasa:
Hello,

I just wanted to point out that the `--extra-files <file...>` command is a bit unintuitive, I had to look it up on Github.  A better way would be  `--extra-files <file> <dest>`.

Prince213:
To anyone who is looking at this, #316 (or #325) has been merged and provides a solution to the problem.

clhodapp:
@Prince213 Unless I'm missing something, it's still counter-intuitive because you don't get to specify the source and destination, as described in the initial report.

Prince213:
> @Prince213 Unless I'm missing something, it's still counter-intuitive because you don't get to specify the source and destination, as described in the initial report.

It's not an exact solution. Instead, the entire path will be copied to the destination, so you can just create the directory structure and put files there. Maybe we can add another option for individual files.

bogorad:
I'm trying to copy my local `/persist/etc/*` to a new remote. zfs filesystem is created, and `tank/persist` is a mount on the new system. this command has no effect. maybe it's because `tank/persist` isn't in its proper place? what's the workaround?

```
 nix run github:nix-community/nixos-anywhere -- --flake .#test root@nixos-live.lan --extra-files /persist/etc
```

bogorad:
I'm trying to copy my local /persist/etc/* to a new remote. a new zfs filesystem is created, and tank/persist is a mount on the new system. this has no effect. maybe it's because tank/persist isn't in its proper place?

```
 nix run github:nix-community/nixos-anywhere -- --flake .#test root@nixos-live.lan --extra-files /persist/etc/creds
```

sedlund:
> I'm trying to copy my local `/persist/*` to a new remote host, into the `/persist` directory. zfs filesystem

Use

> mkdir /target && cp -av /persist /target

Then pass `--extra-files /target`

If /persist and / are on the same zpool this will use copy on write. 

Everything in /target will be copied on top of / on the new system.

bogorad:
> If /persist and / are on the same zpool this will use copy on write.

They are not, and how would that help even if they were? Passing /persist is not the problem.

> Everything in /target will be copied on top of / on the new system.

well, that doesn't solve my problem - I need contents of local /persist/* (which is a dataset on zfs) to be copied to remote /persist/ (which is also a dataset on zfs). The problem is that `tar`, which you switched to from ssh/scp, can't overwrite existing directories.

[...]

PR #440, opened 2024-12-19
docs(extra-files) add howto

sedlund:
Add a howto for `--extra-files`

#283

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -36,8 +36,8 @@ Options:
 * --stop-after-disko
   exit after disko formatting, you can then proceed to install manually or some other way
 * --extra-files <path>
-  path to a directory to copy into the root of the new nixos installation.
-  Copied files will be owned by root.
+  contents of local <path> are recursively copied to the root (/) of the new NixOS installation. Existing files are overwritten
+  Copied files will be owned by root. See documentation for details.
 * --disk-encryption-keys <remote_path> <local_path>
   copy the contents of the file or pipe in local_path to remote_path in the installer environment,
   after kexec but before installation. Can be repeated.
--- a/docs/howtos/extra-files.md
+++ b/docs/howtos/extra-files.md
@@ -1,0 +1,106 @@
+# Copying files to the new installation
+
+The `--extra-files <path>` option allows copying files to the target host after
+installation.
+
+The contents of the `<path>` is recursively copied and overwrites the targets
+root (/). The contents _must_ be in a structure and permissioned as it should be
+on the target.
+
+In this way, there is no need to repeatedly pass arguments (eg: a fictional
+argument: `--copy <source> <dest>`) to `nixos-anywhere` to complete the intended
+outcome.
+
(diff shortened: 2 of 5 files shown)


docs/cli.md
```diff
@@ -36,8 +36,8 @@ Options:
 * --stop-after-disko
   exit after disko formatting, you can then proceed to install manually or some other way
 * --extra-files <path>
-  path to a directory to copy into the root of the new nixos installation.
-  Copied files will be owned by root.
+  contents of local <path> are recursively copied and overwrites root (/) of the new NixOS installation.
```
phaer, line 39:
```suggestion
  contents of local <path> are recursively copied to the root (/) of the new NixOS installation. Existing files are overwritten
```

docs/howtos/extra-files.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Ownership
+
+The new system may have differing UNIX user and group id's for users created
+during installation.
+
+When the files are extracted on the remote the copied data will be owned by
+root.
+
+### Soft links
```
phaer, line 73: Nit: I think the term "symlinks" / "symbolic links" has won and woudl use that here

docs/howtos/extra-files.md
```diff
@@ -0,0 +1,118 @@
...
+
+You want `/etc/ssh/ssh_host_*` and `/persist` from the local system on the
+target.  The `<path>` contents will look like this:
+
+```console
+$ cd /tmp
+$ root=$(mktemp -d)
+$ sudo cp --verbose --archive --link --parents /etc/ssh/ssh_host_* ${root}
+$ cp --verbose --archive --link /persist ${root} # --parents not needed for a full path
+$ lsd --tree --long ${root}
```
phaer, line 30: Maybe remove the lsd call and just post the output as an example layout in another source block? Using non-standard tools in a howto without explaining/installing them could be confusing?

docs/howtos/extra-files.md
```diff
@@ -0,0 +1,118 @@
...
+```sh
+cd /tmp
+root=$(mktemp -d)
+cp --verbose --archive --link --parents /etc/ssh/ssh_host* ${root}
+```
+
+`--parents` will create the directory structure of the source at the
+destination.
+
+### Copy on Write (COW) file systems
```
phaer, line 104: Maybe I am missing something, but this is about space efficiency on the host you are installing from? In a context where the files often don't exceed a few kB?

Love the effort and examples, thank you! But would maybe drop this COW section completely

sedlund: wrt space, I was thinking about what would be in an impermanence setup like listed here:

https://github.com/nix-community/impermanence/blob/c6323585fa0035d780e3d8906eb1b24b65d19a48/README.org?plain=1#L398-L405

but the same case is covered by using hard links, so removed the COW section for brevity

sedlund:
Appreciate the feedback!  Let me know if everything is addressed and I will squash.

phaer:
Thank you!

PR #441, opened 2024-12-20
support for --argstr in the nix-build tf module

noteed:


Mic92:
What is the advantage over special args? Do you have an example?

noteed:
> What is the advantage over special args? Do you have an example?

Special args are for the flake case, and I'm in a project that is not using flakes.

PR #442, opened 2024-12-21
use installer for substituting

Mic92:
Also in most cases our installer will likely not be build from the same nixpkgs revision as our target system, this will still speed up things signifanctly when we don't have to upload drv files twice for `--build-on-remote`.
It might also help people that have their own installer build from the same nixpkgs revision as their target system.


Mic92:
I need to move this to the installer instead.

PR #444, opened 2024-12-22
add option to set ownership of extraFiles

costowell:
I was in need to setting ownership for some extra files and will probably need it more going forward so I added a `--chown` option and documentation too.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -49,7 +49,10 @@ Options:
   copy over existing /etc/ssh/ssh_host_* host keys to the installation
 * --extra-files <path>
   contents of local <path> are recursively copied to the root (/) of the new NixOS installation. Existing files are overwritten
-  Copied files will be owned by root. See documentation for details.
+  Copied files will be owned by root unless specified by --chown option. See documentation for details.
+* --chown <path> <ownership>
+  change ownership of <path> recursively. Recommended to use uid:gid as opposed to username:groupname for ownership.
+  Option can be specified more than once.
 * --disk-encryption-keys <remote_path> <local_path>
   copy the contents of the file or pipe in local_path to remote_path in the installer environment,
   after kexec but before installation. Can be repeated.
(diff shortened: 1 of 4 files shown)


docs/reference.md
```diff
@@ -49,7 +49,10 @@ Options:
   copy over existing /etc/ssh/ssh_host_* host keys to the installation
 * --extra-files <path>
   contents of local <path> are recursively copied to the root (/) of the new NixOS installation. Existing files are overwritten
-  Copied files will be owned by root. See documentation for details.
+  Copied files will be owned by root unless specified by --chown option. See documentation for details.
+* --chown <path> <ownership>
```
Mic92, line 53: Can you add this parameter to our existing tests?

costowell: Done!

Mic92:
@mergify queue

tests/from-nixos.nix
```diff
@@ -62,6 +67,10 @@
       assert "value" == content, f"secret does not have expected value: {content}"
       ssh_key_content = new_machine.succeed(f"cat {ssh_key_path}").strip()
       assert ssh_key_content in ssh_key_output, "SSH host identity changed"
+      priv_key_perms = new_machine.succeed("stat -c %a /home/user/.ssh/").strip()
+      assert priv_key_perms == "600", f"unexpected permissions for private key: {priv_key_perms}"
```
costowell, line 71: Sorry for the long wait. Started this PR near when school started back up for me.

This is the test which is failing. The [docs](https://github.com/nix-community/nixos-anywhere/blob/main/docs/howtos/extra-files.md) say that permissions are preserved, but I don't think that is true. I don't think my PR is the culprit, just the first time a test was written for permissions.

Should I make another PR to fix the permissions problems? @Mic92

costowell:
Ignore my previous message, everything looks good! Let me know if anything more needs to be done.

costowell:
I'm a little unsure as to what to do. One of the tests failed due to a kernel panic, and I can't imagine that's due to my PR. Any comments or pointers would be greatly appreciated.

Mic92:
@mergify queue

Mic92:
> I'm a little unsure as to what to do. One of the tests failed due to a kernel panic, and I can't imagine that's due to my PR. Any comments or pointers would be greatly appreciated.

No action required. This is general flakiness because we are close to the memory limit for some reason.

Mic92:
Tried to mitigate it here: https://github.com/nix-community/nixos-anywhere/pull/473

PR #447, opened 2025-01-07
Don't run preSwitchChecks during install

phaer:
NixOS includes an option called `system.preSwitchChecks` since https://github.com/NixOS/nixpkgs/commit/6e192c4489ea44c48876faa11060798fb9502146.

As we are installing a new system here, we do not care too much about the systems state pre switching and therefore disable those checks. 

This change is motivated by https://github.com/nix-community/srvos/pull/591#issuecomment-2575221123

untested, but this might help @Kranzes?

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -614,7 +614,7 @@ if [ ${copyHostKeys-n} = "y" ]; then
     cp -a "\$p" "/mnt/\$p"
   done
 fi
-nixos-install --no-root-passwd --no-channel-copy --system "$nixosSystem"
+NIXOS_NO_CHECK=1 nixos-install --no-root-passwd --no-channel-copy --system "$nixosSystem"
 if [[ ${phases[reboot]} == 1 ]]; then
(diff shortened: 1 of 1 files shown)


phaer:
Happy about alternative PRs! If you want to convince me, I'd need more of an argument as to why you think so.

Srvos already ships two mechanism by which to skip this check.

EXPECTED_HOSTNAME, which requires the target hostname that we don't have readily available.
And NIXOS_NO_CHECK which is used here as I couldn't come up with a preSwitchCheck that you would *want* to run during first install.

Kranzes:
> If you want to convince me, I'd need more of an argument as to why you think so.

You won. I don't have the energy LOL. I will test this PR later.

zowoq:
PR to improve installer detection in https://github.com/nix-community/srvos/pull/601.

Mic92:
@phaer do you still think this PR is needed?

phaer:
@Mic92 "needed" might be a too strong word. Still think it's probably closer to what users wants and therefore useful, no?

Alternatively could try to fix that upstream

PR #452, opened 2025-01-13
feat: Use `identity_file` as a deployment key

a-kenji:
Will now use the provided `-i` (identity_file) as a stable deployment key - if specified - rather than always generating a random ephemeral SSH key.

This is primarily useful for the following reasons:
- error recovery
- controlled phases

*Error Recovery*

If `nixos-anywhere` stops with an error, we might want to connect remotely to the deployed machine. With the ephemeral key this might not be possible, because we don't necessarily have access to it anymore.

*Controlled Phases*

It is already possible to control which phases should be run. With an ephemeral key outside our control we are not able to resume the phases with a second `nixos-anywhere` invocation, because a different deployment key will now be generated.


phaer (approved):
Code LGTM, Thanks!

Lassulus:
@mergify queue

Issue #450, opened 2025-01-10
Using swap for install


pmaterer:
I'm using nixos-anywhere to install NixOS on a machine with 1GB of memory. During the Disko phase the `/nix/store` runs out of space (`tmpfs` filesystem running in memory). So I've taken to using a larger (30GB) external disk for swap. And this works, but when it gets to the reboot phase it fails. This is because I'm configuring zpools using Disko, and nixos-anywhere runs [`swapoff -a`](https://github.com/nix-community/nixos-anywhere/blob/main/src/nixos-anywhere.sh#L622) if it detects zpools. I understand this was added because of #420  (@TilCreator is using swap on a zpool) which makes sense... but maybe we can add a flag to disable this behavior. This flag would be useful when you are using swap for the install and not using swap on a zpool.

pmaterer:
FWIW I immediately get an "Out of memory" error after swap is turned off and my zpools do not get exported.

useranon350:
Swap on zpool isn't really supported by upstream ZFS BTW. It's known to have deadlock issues when memory pressure is high, or when certain features like hibernate and suspend are used.

PR #421, opened 2024-11-05
add swapoff to reboot phase

TilCreator:
fixes #420

now tested!

PR #454, opened 2025-01-14
Add swapoff flag

pmaterer:
This is a fix for #450.

Adds a flag `--swapoff` which enables the features added in #421. 

So now, by default, `swapoff -a` will not run during the reboot phase. Only when you have a ZFS swap device would you need to pass the `--swapoff` flag. This way users who are using a swap device during the install won't have things break.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -53,4 +53,6 @@ Options:
   build the closure on the remote machine instead of locally and copy-closuring it
 * --vm-test
   build the system and test the disk configuration inside a VM without installing it to the target.
+* --swapoff
+  turn off swap devices during reboot phase. Enable this if creating a ZFS swap device so it is exported.
 ```
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -12,6 +12,7 @@ diskoMode="disko"
 nixosSystem=""
 extraFiles=""
 vmTest="n"
+swapoff="n"
 nixOptions=(
   --extra-experimental-features 'nix-command flakes'
   "--no-write-lock-file"
@@ -117,6 +118,8 @@ Options:
 * --generate-hardware-config nixos-facter|nixos-generate-config <path>
   generate a hardware-configuration.nix file using the specified backend and write it to the specified path.
   The backend can be either 'nixos-facter' or 'nixos-generate-config'.
+* --swapoff
+  turn off swap devices during reboot phase. Enable this if creating a ZFS swap device so it is exported.
 * --phases
   comma separated list of phases to run. Default is: kexec,disko,install,reboot
   kexec: kexec into the nixos installer
@@ -288,6 +291,9 @@ parseArgs() {
       ;;
     --vm-test)
       vmTest=y
+      ;;
+    --swapoff)
+      swapoff=y
       ;;
     *)
       if [[ -z ${sshConnection} ]]; then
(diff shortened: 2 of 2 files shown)


src/nixos-anywhere.sh
```diff
@@ -619,7 +625,9 @@ if [[ ${phases[reboot]} == 1 ]]; then
   if command -v zpool >/dev/null && [ "\$(zpool list)" != "no pools available" ]; then
     # we always want to export the zfs pools so people can boot from it without force import
     umount -Rv /mnt/
-    swapoff -a
+    if [[ ${swapoff} == "y" ]]; then
+      swapoff -a
```
Mic92, line 629: We can probably just do it by default. I currently don't see a disadvantage if our next step is to reboot.

docs/cli.md
```diff
@@ -53,4 +53,6 @@ Options:
   build the closure on the remote machine instead of locally and copy-closuring it
 * --vm-test
   build the system and test the disk configuration inside a VM without installing it to the target.
+* --swapoff
+  turn off swap devices during reboot phase. Enable this if creating a ZFS swap device so it is exported.
```
Mic92, line 57: Than we wouldn't need a flag.

src/nixos-anywhere.sh
```diff
@@ -619,7 +625,9 @@ if [[ ${phases[reboot]} == 1 ]]; then
...
+    fi
```
Mic92, line 630:
```suggestion
    # in case our zpool has a swap device
    swapoff -a
```

src/nixos-anywhere.sh
```diff
@@ -289,6 +292,9 @@ parseArgs() {
     --vm-test)
       vmTest=y
       ;;
+    --swapoff)
+      swapoff=y
+      ;;
     *)
```
Mic92, line 298:
```suggestion
    *)
```

src/nixos-anywhere.sh
```diff
@@ -117,6 +118,8 @@ Options:
 * --generate-hardware-config nixos-facter|nixos-generate-config <path>
   generate a hardware-configuration.nix file using the specified backend and write it to the specified path.
   The backend can be either 'nixos-facter' or 'nixos-generate-config'.
+* --swapoff
+  turn off swap devices during reboot phase. Enable this if creating a ZFS swap device so it is exported.
 * --phases
```
Mic92, line 123:
```suggestion
  The backend can be either 'nixos-facter' or 'nixos-generate-config'.
* --phases
```

docs/cli.md
Mic92, line 57:
```suggestion
```

src/nixos-anywhere.sh
Mic92, line 629: What I don't like about this change is that it breaks the export the when people have swap on ZFS.
People shouldn't need to have adding a flag to make nixos-anywhere work.

Mic92: I think something like this is more sensible:

```suggestion
    swapon --show --noheadings | awk '/^\/dev\/zd/ {print $1}' | while read device; do
      swapoff "$device"
    done
```

You would need to have your system not have swap in zfs. However you need to do this anyway because with 1GB you would run into deadlock issues with zfs if you would try to have it in the pool.

Mic92:
@pmaterer is this still required after: https://github.com/nix-community/disko/blob/76c0a6dba345490508f36c1aa3c7ba5b6b460989/disk-deactivate/zfs-swap-deactivate.jq#L4
?

Mic92:
In short disko now will turn of zfs-based swap.

pmaterer:
> @pmaterer In disko itself we now have this to turn of swap that is based on zfs: https://github.com/nix-community/disko/blob/76c0a6dba345490508f36c1aa3c7ba5b6b460989/disk-deactivate/zfs-swap-deactivate.jq#L4 ?

Great! I will try it out. I think we can close this then.

PR #456, opened 2025-01-20
--debug enables --show-trace on nix build commands

Qubasa:
Just a little debugging improvement

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -7,6 +7,7 @@ flakeAttr=""
 kexecUrl=""
 kexecExtraFlags=""
 enableDebug=""
+nixBuildFlags=""
 diskoScript=""
 diskoMode="disko"
 nixosSystem=""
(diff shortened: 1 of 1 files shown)


src/nixos-anywhere.sh
```diff
@@ -7,6 +7,7 @@ flakeAttr=""
 kexecUrl=""
 kexecExtraFlags=""
 enableDebug=""
+nixBuildFlags=""
```
Mic92, line 10: Actually I would prefer a separate `--show-trace` flag here. For the majority of users `--show-trace` is not very useful for NixOS and just make things less readable.

Qubasa: True, added it to a separate flag

dasJ:
Was this tested?

```
++ nix build --print-out-paths --no-link ' --show-trace' --extra-experimental-features 'nix-command flakes' --no-write-lock-file -L '.#nixosConfigurations."test".config.system.build.diskoScript'
error: getting status of '/home/janne/ --show-trace': No such file or directory
```

PR #457, opened 2025-02-01
docs: Add a reference to clan to quickstart

a-kenji:
`clan` is a deployment framework that also makes use of `nixos-anywhere`.

References:
- https://clan.lol/
- https://docs.clan.lol/getting-started/deploy


Mic92:
@mergify queue

PR #460, opened 2025-02-05
feat: Add package list metadata to Terraform build output via nvd

oneingan:
### **Changes**  
This PR enhances the Nix build script to:  
1. Run `nvd list` on the derivation output to collect installed packages.  
2. Format the package list as a JSON-compatible string for Terraform compatibility.  
3. Return a flat JSON object with `out` (build path) and `packages` (serialized JSON array).  

**Example Output**:  
```json
{
  "out": "/nix/store/...-nixos-system",
  "packages": "[{\"name\":\"acl\",\"versions\":\"2.3.2 x2\"},...]"
}
```

### **Notes**  
- Requires `nvd` installed, but if not returns an empty list
- Maintains backward compatibility – existing users see no breaking changes.

### DIFF
--- a/terraform/nix-build/nix-build.sh
+++ b/terraform/nix-build/nix-build.sh
@@ -39,5 +39,20 @@ else
   # inject `special_args` into nixos config's `specialArgs`
   # shellcheck disable=SC2086
   out=$(nix build --no-link --json ${options} --expr "${nix_expr}" "${config_attribute}")
+  output_path=$(echo "$out" | jq -r '.[].outputs.out')
+  packages_json=$(nvd list --selected --root "$output_path" | awk '
+  BEGIN { first=1; printf "[" }
+  {
+    if (match($0, /^\[I\*\]\s+#([0-9]+)\s+([^ ]+)\s+(.*)$/, arr)) {
+      if (!first) { printf "," }
+      first=0
+      printf "{\"name\": \"%s\", \"versions\": \"%s\"}", arr[2], arr[3]
+    }
+  }
+  END { print "]" }
+  ')
 fi
-printf '%s' "$out" | jq -c '.[].outputs'
+jq -n \
+  --arg out "$output_path" \
+  --arg packages "$packages_json" \
+  '{ "out": $out, "packages": $packages }'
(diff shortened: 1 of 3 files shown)


terraform/nixos-rebuild/variables.tf
```diff
@@ -3,6 +3,12 @@ variable "nixos_system" {
   description = "The nixos system to deploy"
 }
 
+variable "nixos_system_packages" {
```
Mic92, line 6: Mhm. I don't think i want this type of adhoc configuration and re-implement all kinds of NixOS configuration.
If you really want to configure this from terraform, I would suggest to `git add` files to the repository with terraform or use specialArgs to inject information into nixOS and have your own modules to process it in your module.

Mic92: See https://github.com/nix-community/nixos-anywhere/pull/418 for details.

Mic92: Ah. I think I miss understood the feature.

terraform/nix-build/nix-build.sh
```diff
@@ -39,5 +39,20 @@ else
   # inject `special_args` into nixos config's `specialArgs`
   # shellcheck disable=SC2086
   out=$(nix build --no-link --json ${options} --expr "${nix_expr}" "${config_attribute}")
+  output_path=$(echo "$out" | jq -r '.[].outputs.out')
+  packages_json=$(nvd list --selected --root "$output_path" | awk '
```
Mic92, line 43: Maybe add some conditional if nvd is not present?

```
if command -v nvd; then
fi
```

terraform/nixos-rebuild/main.tf
```diff
@@ -1,6 +1,7 @@
 resource "null_resource" "nixos-rebuild" {
   triggers = {
     store_path = var.nixos_system
+    packages_list = var.nixos_system_packages
```
Mic92, line 4: I am not quite getting, why we need this trigger? Is the store_path not enough?

oneingan: If I'm not mistaken, you need the for `package_list` diff to be displayed during the plan, though I could be wrong.

Mic92: I see. Actually the terraform option description is a bit misleading in this case, which was also the initial source of my confusion. Could you change that and also add to `nixos-rebuild.md` how to wire things up i.e. use nix-build output and feed it into the nixos-rebuild module?

PR #466, opened 2025-02-08
Revert "Add --show-trace flag, to show nix build traces"

Lassulus:
This reverts commit 73b77b1c5e691c30b226bf2349ca8708c68c3f67.

if not passing --show-trace this builds the flakes default argument (because the flags are just an empty string ""). So installing fails if this default package is not defined or fails to substitute on the target


Mic92:
Can you test https://github.com/nix-community/nixos-anywhere/pull/465?

Mic92:
Apparently the other one fixes the issue.

PR #475, opened 2025-02-27
Back out "Don't run preSwitchChecks during install"

Enzime:
This backs out commit be2d9d752660c95bb5232cd9829e150d05a90f25.

This backs out of #447

Now that https://github.com/nix-community/srvos/pull/591#issuecomment-2575221123 is fixed upstream, I think it would be preferable to run the pre-switch checks before installing to match the default behaviour of `nixos-install` and I usually think of `nixos-install` as `nixos-rebuild switch` with a few extra steps like installing a bootloader.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -719,7 +719,7 @@ if [ ${copyHostKeys-n} = "y" ]; then
     cp -a "\$p" "/mnt/\$p"
   done
 fi
-NIXOS_NO_CHECK=1 nixos-install --no-root-passwd --no-channel-copy --system "$nixosSystem"
+nixos-install --no-root-passwd --no-channel-copy --system "$nixosSystem"
 if [[ ${phases[reboot]} == 1 ]]; then
   if command -v zpool >/dev/null && [ "\$(zpool list)" != "no pools available" ]; then
     # we always want to export the zfs pools so people can boot from it without force import


Enzime:
In the original PR adding the pre-switch checks, it explicitly mentions they are meant to run during installation (prior to the installation of the bootloader) https://github.com/NixOS/nixpkgs/pull/236375

phaer (changes requested):
The specific srvos check you are linking to was fixed before #447 was merged here.

As written there: we don't switch from one system to another during install, because in the case where it's an empty disk, there's just no system to speak of before the switch.
And if you use it because the original system is too broken to do a nixos-rebuild inside, preSwitchChecks most probably won't neither pass nor help.

So I'd like to see a more practical argument than just a mental model (which isn't entirely correct imo) before a possible revert?

I mean, I don't care too much either way, I just think it's much more practical this way, but we could always make it optional if needed.

phaer:
>  it explicitly mentions they are meant to run during installation (prior to the installation of the bootloader)

The PR description says

> Add an option for shell script fragments that are ran before switching to a new NixOS system configuration (pre installation of bootloader or system activation).

Which I parse as "during switch to new system, but before bootloader is installed". It does not say anything explicitly about the installation of new systems.

Enzime:
@Mic92 was there a specific issue you ran into recently that made you merge #447 recently?

I think that we should match upstream rather than adding an extra flag to diverge our behaviour. In my opinion, pre-switch checks should be able to handle when there's no system/state yet.

From `nixos-install`'s source code:

> Switch to the new system configuration.

https://github.com/NixOS/nixpkgs/blob/82e9fc276ac051cc6221d0d1a35368ca5746492d/pkgs/by-name/ni/nixos-install/nixos-install.sh#L239

Mic92:
I had no particular issue. It was more about the fact that many probably don't test installation very often for their preSwitch checks.

phaer:
to be honest, I don't think it's good style to merge this and cut a new release right after, but before a rough consensus is reached while ignoring my arguments.

I do agree that this should be fixed upstream before 25.05. And I am fine with keeping it optional. But not blocking installation should be a priority from my POV and this did block installations in the past and will do so again.
Even with an upstream change we'll have to wait for the next stable release to ensure it doesn't block anything. 

> I think that we should match upstream rather than adding an extra flag to diverge our behaviour. In my opinion, pre-switch checks should be able to handle when there's no system/state yet.

We are not "adding an extra flag", we are setting it. There's no material difference between doing that and i.e. enabling support for flakes which is also diverging from upstreams default behaviour. But it's an opinionated choice one can make.

PR #478, opened 2025-02-27
Allow skipping system.preSwitchChecks...

phaer:
as they might be written for switching from one generation to another, without initial installation in mind and failing checks would prevent successful installations.

This is a follow up on both #447 and #475, after a bit of discussion on matrix.
I still think that skipping those checks in nixos-install by default would have been a better default, but am not sure whether it's worth arguing about that upstream now.
The main reason is that it's at least possible to opt-out of preSwitchChecks with nixos-install in case a check is broken. This isn't possible with nixos-anywhere (before #447 and since #475), but will be after this PR. This is sufficient in order not to block installations in the presence of broken checks.

Trade-off is yet another obscure flag  ;)

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -719,6 +719,10 @@ if [ ${copyHostKeys-n} = "y" ]; then
     cp -a "\$p" "/mnt/\$p"
   done
 fi
+# https://stackoverflow.com/a/13864829
+if [ ! -z ${NIXOS_NO_CHECK+0} ]; then
+  export NIXOS_NO_CHECK
+fi
 nixos-install --no-root-passwd --no-channel-copy --system "$nixosSystem"
 if [[ ${phases[reboot]} == 1 ]]; then
   if command -v zpool >/dev/null && [ "\$(zpool list)" != "no pools available" ]; then


Enzime (approved):
I don’t mind having another flag, but if you pass `NIXOS_NO_CHECK=1` to `nixos-anywhere` it should skip the checks just like `nixos-install` I believe

I think the root cause of why people have even had issues is because the error message for the pre-switch checks doesn’t mention `NIXOS_NO_CHECK`

phaer:
> I don’t mind having another flag, but if you pass NIXOS_NO_CHECK=1 to nixos-anywhere it should skip the checks just like nixos-install I believe

I thought no because SSH doesn't pass env vars through and nixos-install runs on the remote side ofc.
But for some reason, I currently don't get a failing preSwitchCheck to work in our test framework to properly reproduce atm and need to jump soon - will continue on that later.

src/nixos-anywhere.sh
```diff
@@ -719,6 +725,9 @@ if [ ${copyHostKeys-n} = "y" ]; then
     cp -a "\$p" "/mnt/\$p"
   done
 fi
+if [ ${skipPreSwitchChecks-n} = y ]; then
+  export NIXOS_NO_CHECK=1
```
Mic92, line 729: if ssh doesn't allow to pass the environment flag, we can just forward it manually here instead of yet another cli flag.

phaer: You can verify that behavior with any SSH server. Maintainer edits are open :)

phaer: Just looked this again, because I didn't want to leave this around for too long:

Upstream (both the classic nixos-rebuild as well as the -ng variant) does check if it equals 1, not just if it's set. So using -z would diverge from upstream here.

Enzime: I've updated this to just check if `NIXOS_NO_CHECK` is set, if it is set then just pass that variable by exporting it to `nixos-install`, that way we don't need to care what the value of this variable is

PR #489, opened 2025-03-19
docs(no-os): simpler and more complete

yajo:
- Remind user to connect to WiFi.
- Remove the IP section. Just connect to localhost and it will work always.
- Remind to add experimental features to CLI.
- Add `--build-on remote`, or most likely the live system will run out of disk space (which is actually RAM at that moment).

@moduon MT-9339

### DIFF
--- a/docs/howtos/no-os.md
+++ b/docs/howtos/no-os.md
@@ -18,53 +18,21 @@ If an installer is detected, `nixos-anywhere` will not attempt to `kexec` into
 its own image. This is particularly useful for targets that don't have enough
 RAM for `kexec` or don't support `kexec`.
 
+Often you will need some kind of network connectivity before installing NixOS.
+Use the live system to connect to some network.
+
 NixOS starts an SSH server on the installer by default, but you need to set a
 password in order to access it. To set a password for the `nixos` user, run the
 following command in a terminal on the NixOS machine:
 
-```
+```shell
 passwd
 ```
 
-If you don't know the IP address of the installer on your network, you can find
-it by running the following command:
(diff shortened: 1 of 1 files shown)


docs/howtos/no-os.md
```diff
@@ -18,53 +18,21 @@ If an installer is detected, `nixos-anywhere` will not attempt to `kexec` into
...
-```
-ssh -v nixos@fec0::5054:ff:fe12:3456
-```
-
-You can then use the IP address to run `nixos-anywhere` like this:
-
-```
-nix run github:nix-community/nixos-anywhere -- --flake '.#myconfig' --target-host nixos@fec0::5054:ff:fe12:3456
+```shell
+nix --extra-experimental-features 'nix-command flakes' run github:nix-community/nixos-anywhere -- --flake '.#myconfig' --target-host nixos@localhost --build-on remote
```
Mic92, line 35: This way works well if you have a public configuration. However if you don't it's own challenge to clone the repository. So I think we should maybe describe both method.

yajo: But the previous version didn't explain how to clone a repo. What do you mean?

Mic92: Let's say your have a private configuration with private flake inputs it's more likely you have the necessary credentials available on an already set up machine vs. an installer image.

yajo: OK, so you mean that these instructions were meant to be executed from a different PC within the same network, that is targeting the new PC that was booted with the live image, right?

That wasn't obvious to me when reading the docs. Let me fix now.

PR #490, opened 2025-03-19
fix arguments check in deploy.sh

asymmetric:
The previous commit added an argument but didn't increase the check on their number.


phaer (approved):
Haven't tested it, but the fact that you two did and that the number of args wasn't updated in #487 and that it's easy to count makes me confident that this is good to merge ;)

cc @r-vdp

r-vdp:
*facepalm*, sorry about that!

PR #504, opened 2025-04-18
Actually reboot if the reboot phase is specified

jfly:
Previously, we'd only reboot if both the install *and* the reboot phases are specified.


Lassulus (approved):

Mic92:
@jfly had to rebase. Does the new diff looks good?

jfly:
> @jfly had to rebase. Does the new diff looks good?

LGTM!

PR #522, opened 2025-05-22
nixos-anywhere: don't prompt SSH agent after deployment key installed

Enzime:


src/nixos-anywhere.sh
```diff
@@ -559,7 +552,7 @@ checkBuildLocally() {
     -L \
     "${nixOptions[@]}" \
     --expr \
-    "derivation { system = \"$system\"; name = \"env-$entropy\"; builder = \"/bin/sh\"; args = [ \"-c\" \"echo > \$out\" ]; }"; then
+    "derivation { system = \"$machineSystem\"; name = \"env-$entropy\"; builder = \"/bin/sh\"; args = [ \"-c\" \"echo > \$out\" ]; }"; then
```
sedlund, line 555: Wouldn't this fail as `$machineSystem` would be unset when called from here:
https://github.com/nix-community/nixos-anywhere/blob/1e2882c902322f1fd5992da0f5674c70ad8b133b/src/nixos-anywhere.sh#L787-L789

Enzime: It gets set at the start of the `checkBuildLocally` function:

https://github.com/nix-community/nixos-anywhere/blob/669a9b0ba1fcabf28dac82693800ad86bb7771f2/src/nixos-anywhere.sh#L521-L536

sedlund: but $# would be 0 when called without args from line 788 - so would not be set?

Enzime: It would go through the `elif` and `else` clauses that would set it

PR #527, opened 2025-05-30
runKexec: reduce number of ssh command needed for kexec

Mic92:
if the user has a sudo password set, we were requiring multiple times to enter the sudo password. This now runs everything in one command to reduce the number of sudo invocation.


pinpox (approved):
Approved as discussed

PR #550, opened 2025-06-09
make installer with sudo-only access work and re-enable ssh-ng

Mic92:
Now that our nixos installers should have a fixed ssh-ng implementation,
it should be safe to re-enable it.

On top we now test the use-case of an installer that doesn't need kexec
but sudo explicitly.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -64,7 +64,7 @@ mkdir -p "$tempDir"
 
 declare -A diskEncryptionKeys=()
 declare -A extraFilesOwnership=()
-declare -a nixCopyOptions=()
+declare -a nixCopyOptions=(--no-check-sigs)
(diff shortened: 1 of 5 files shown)


src/nixos-anywhere.sh
```diff
@@ -898,13 +943,6 @@ main() {
     fi
   fi
 
-  # Installation will fail if non-root user is used for installer.
-  # Switch to root user by copying authorized_keys.
-  if [[ ${isInstaller} == "y" ]] && [[ ${sshUser} != "root" ]]; then
-    # Allow copy to fail if authorized_keys does not exist, like if using /etc/ssh/authorized_keys.d/
-    runSsh "${maybeSudo} mkdir -p /root/.ssh; ${maybeSudo} cp ~/.ssh/authorized_keys /root/.ssh || true"
```
Mic92, line 905: This scripts makes the assumption that the ssh key in /root are respected by the sshd and that the root user is allowed to login via ssh.

Mic92:
This should also simplify users who want to turn a Linux distribution without root login into a NixOS installer with a script.

PR #554, opened 2025-06-13
nixos-anywhere: set up `substituters` from target machine in installer

Enzime:


Enzime:
This PR adds a runtime dependency on `jq`, I wonder if there's a good way around it or should we just add it as a dependency in the derivation

PR #566, opened 2025-07-18
limited-ram: document that having a swap partition will help

tomfitzhenry:
Follow up from https://github.com/nix-community/nixos-images/pull/370 to mention in the docs to use a swap partition if on a low RAM machine.

### DIFF
--- a/docs/howtos/limited-ram.md
+++ b/docs/howtos/limited-ram.md
@@ -1,9 +1,12 @@
 # Kexec on Systems with Limited RAM
 
 When working with nixos-anywhere on systems with limited RAM (around 1GB), you
-can use the `--no-disko-deps` option to reduce memory usage during installation.
+can:
 
-## How it works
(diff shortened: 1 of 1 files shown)


docs/howtos/limited-ram.md
```diff
@@ -1,9 +1,12 @@
 # Kexec on Systems with Limited RAM
 
 When working with nixos-anywhere on systems with limited RAM (around 1GB), you
-can use the `--no-disko-deps` option to reduce memory usage during installation.
+can:
 
-## How it works
+* use the `--no-disko-deps` option to reduce memory usage during installation.
+* ensure the disk has a swap partition
```
a-kenji, line 7: I am not too happy with this exact wording. It also seems to contradict with your commit message a little.

To me
 > ensure the disk has a swap partition

Reads like you *need* to have a swap partition, not that a swap partition might or will help.
It also can be read as `--no-disko-deps` requires a swap partition.

Issue #567, opened 2025-07-19
sshHost not defined in runKexec


b-naber:
### Prerequisites

- [x] I have updated to the latest version using `nix run --refresh github:nix-community/nixos-anywhere`
- [x] I have reproduced the issue with the `--debug` flag
- [x] I have searched existing issues to make sure this isn't a duplicate

### Bug Description

I'm getting the following when using the terraform module (but this applies to nixos-anywhere itself too):

```
module.deploy-pg-archive.module.install.null_resource.nixos-remote (local-exec): + runSsh -o ConnectTimeout=10 -- exit 0
module.deploy-pg-archive.module.install.null_resource.nixos-remote (local-exec): + ssh -T -o IdentitiesOnly=yes -i /tmp/tmp.Msmquh0UOA/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p 22 root@ -o ConnectTimeout=10 -- exit 0
module.deploy-pg-archive.module.install.null_resource.nixos-remote (local-exec): ssh: Could not resolve hostname : Name or service not known
module.deploy-pg-archive.module.install.null_resource.nixos-remote (local-exec): + sleep 5
```

It looks as if `sshHost`, which was introduced by https://github.com/nix-community/nixos-anywhere/pull/562, is [not defined inside `runKexec`](https://github.com/nix-community/nixos-anywhere/blob/d00d5b73af8b777561b83b62519c315cfd7a69b7/src/nixos-anywhere.sh#L698)

### Steps to Reproduce

-

### Debug Logs

```shell
-
```

### Command Used

-

### Target System

Cloud server (AWS, DigitalOcean, etc.)

### NixOS Version

23.11

### Environment Information

- Host OS: 
- Nix version:
- Target architecture:

### Configuration Files

```nix

```

Meallia:
I'll have a look, but it would be easier with the nixos-anywhere debug logs @b-naber   ;-)

b-naber:
`OpenSSH_8.9p1 Ubuntu-3ubuntu0.13, OpenSSL 3.0.2 15 Mar 2022`

Meallia:
Thanks.

After a quick leap in the past thanks to nixhub.io, It looks like #562 introduced a dependency on openssh >= 9.2.

I think the easiest fix would be to remove the `user@` part from the SSH connection instead of relying on openssh to do it for us. I'll open a quick MR for that.

a-kenji:
[...]

PR #568, opened 2025-07-20
Remove dependency on openssh >= 9.2

Meallia:
#562 Introduced a dependency on openssh >= 9.2 as discovered in #567 

Instead of relying `ssh -G` to give the hostname field, we can just compute it instead.


a-kenji (approved):
:+1:  cc @Enzime

Issue #410, opened 2024-10-15
docs: cli.md outdated / missing --phases


sedlund:
[docs/cli.md](https://github.com/nix-community/nixos-anywhere/blob/main/docs/cli.md)

Update or deprecate this doc ?

PR #571, opened 2025-07-24
docs: Update the CLI documentation

a-kenji:
Closes: #410

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -8,9 +8,12 @@ Usage: nixos-anywhere [options] [<ssh-host>]
 Options:
 
 * -f, --flake <flake_uri>
-  set the flake to install the system from.
+  set the flake to install the system from. i.e.
+  nixos-anywhere --flake .#mymachine
+  Also supports variants:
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -1,14 +1,19 @@
 # CLI
 
+<!-- `$ nix run . -- --help` -->
```
Enzime, line 3: What's this for?

a-kenji: It's in the commit message. It's to easily keep the cli documentation up to date.

Enzime: Does that mean running the command in the commit message will automatically update this file?

We already have `mdsh` configured under `treefmt` so does that mean our `checks.<system>.treefmt` will now error if it's out of date?

a-kenji: > Does that mean running the command in the commit message will automatically update this file?

Yes, running the command will automatically update the file and keep it up to date.

> We already have mdsh configured under treefmt so does that mean our checks.<system>.treefmt will now error if it's out of date?

No, and we could remove `mdsh` from `treefmt` as it seems unused to me.

PR #572, opened 2025-07-25
chore: Remove `mdsh` from `nix fmt` configuration

a-kenji:
Remove `mdsh` from the `nix fmt` configuration.
Making `nix fmt` and `nix flake check` take a little less resources,
since `mdsh` is currently not used.


Mic92:
> Instead of removing it, could we fix it so it picks up those comment directives?

Ideally, but so far anyone has noticed.

PR #573, opened 2025-07-29
Fix sudo password prompt

Qubasa:
### Problem
- Currently nixos-anywhere fails if you test it with the default ubuntu server image. As it configures a normal user that has sudo with password access to root. 

```bash
TMPDIR=/root/kexec setsid --wait ${maybeSudo} /root/kexec/kexec/run --kexec-extra-flags $(printf '%q ' "$kexecExtraFlags") 
```
the problem / bug is in the line above where we run kexec with sudo inside a setsid terminal, that sudo needs to re-prompt for the password, and doesn't have access to stdin so it fails.

- Another problem is that errors happening in the kexec step are not propagated through to the shell

## My Changes

- Add a `breakpoint` function, it halts execution and gives you an interactive bash shell with all the internal variables in scope

- Reduce log spamming by adding `set +x` to for loops

- Make `remoteCommand` write a shell script that then get's called once with sudo, instead of calling sudo inside the setsid shell

## Tested

I have tested it with the 
- `--kexec ./nixos-kexec-installer-noninteractive-x86_64-linux.tar.gz` flag 
- `--kexec https://gh-v6.com/nix-community/nixos-images/releases/download/nixos-unstable/nixos-kexec-installer-noninteractive-x86_64-linux.tar.gz`
- and without the `--kexec` flag
successfully on an `ubuntu24.04` image with non root user `qubasa@<ip>`

## Summary by CodeRabbit

* **New Features**
  * Interactive debugging breakpoint with a temporary debug environment and shell.
  * Centralized remote logging with automatic retrieval on failures.
* **Enhancements**
  * Clearer SSH command display in debug mode.
  * More reliable kexec execution with unified success/failure handling.
  * Automatic switch to root for post-kexec and installer steps when needed.
  * Smarter tarball handling: prefers local upload; falls back to remote download if unavailable.
  * Fact gathering now includes remote home directory for consistent paths.
  * Broader debug/log improvements across SSH and remote interactions.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -1016,6 +1016,12 @@ main() {
   sshUser=$(echo "$sshSettings" | awk '/^user / { print $2 }')
   sshHost="${sshConnection//*@/}"
 
+  # If kexec phase is not present, we assume kexec has already been run
+  # and change the user to root@<sshHost> for the rest of the script.
+  if [[ ${phases[kexec]} != 1 ]]; then
+    sshConnection="root@${sshHost}"
+  fi
+
   uploadSshKey
 
   importFacts


src/nixos-anywhere.sh
```diff
@@ -643,13 +692,53 @@ runKexec() {
...
   # Define common remote commands template
   local remoteCommandTemplate
-  remoteCommandTemplate="
+
+  # If we need sudo and have a TTY, use a script that can handle password prompts
+  if [[ -n ${maybeSudo} ]] && [[ ${sshTtyParam} == "-t" ]]; then
+    remoteCommandTemplate="
+${enableDebug:+set -x}
+# Create a script that we can run with sudo
+cat > /tmp/kexec-script.sh << 'KEXEC_SCRIPT'
```
Mic92, line 701: mktemp is preferred here because of https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File
we might leak a file but this still I think.

src/nixos-anywhere.sh
```diff
@@ -643,13 +692,53 @@ runKexec() {
...
+#!/bin/bash
+set -eu ${enableDebug}
+rm -rf /root/kexec
+mkdir -p /root/kexec
+cd /root/kexec
+echo \"Downloading kexec tarball (this may take a moment)...\"
+# Execute tar command
+%TAR_COMMAND% && TMPDIR=/root/kexec setsid --wait /root/kexec/kexec/run --kexec-extra-flags $(printf '%q ' "$kexecExtraFlags")
+KEXEC_SCRIPT
+chmod +x /tmp/kexec-script.sh
```
Mic92, line 711: Than we also don't need chmod +x

src/nixos-anywhere.sh
```diff
@@ -643,13 +692,53 @@ runKexec() {
...
+# Run the script and let output flow naturally  
+${maybeSudo} /tmp/kexec-script.sh 2>&1 | tee /tmp/kexec-output.log || true
```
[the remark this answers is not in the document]
Qubasa: @Mic92 Ah because of a potential race condition with an attacker you are right, I didn't think of that. 

The output log is static because then we can get the logs from the host easily without needing to back communicate the path, but I can do something about it

src/nixos-anywhere.sh
Qubasa: Moved it to a dynamic temp dir

src/nixos-anywhere.sh
```diff
@@ -643,13 +693,55 @@ runKexec() {
...
+
+  # If we need sudo and have a TTY, use a script that can handle password prompts
+  if [[ -n ${maybeSudo} ]] && [[ ${sshTtyParam} == "-t" ]]; then
+    remoteCommandTemplate="
+${enableDebug:+set -x}
+# Create a script that we can run with sudo
+kexec_script_tmp=$(mktemp /tmp/kexec-script.XXXXXX.sh)
+trap 'rm -f \"\$kexec_script_tmp\"' EXIT
+cat > \"\$kexec_script_tmp\" << 'KEXEC_SCRIPT'
+#!/usr/bin/env bash
```
Mic92, line 705:
```suggestion
```

src/nixos-anywhere.sh
```diff
@@ -657,14 +716,63 @@ runKexec() {
...
+rm -rf /root/kexec
+mkdir -p /root/kexec
+cd /root/kexec
+echo 'Downloading kexec tarball (this may take a moment)...'
+# Execute tar command
+%TAR_COMMAND% && TMPDIR=/root/kexec setsid --wait /root/kexec/kexec/run --kexec-extra-flags $(printf '%q ' "$kexecExtraFlags")
+KEXEC_SCRIPT
+
+# Run the script and let output flow naturally  
+${maybeSudo} bash \"\$kexec_script_tmp\" 2>&1 | tee /tmp/kexec-output.log || true
```
Mic92, line 767:
```suggestion
${maybeSudo} bash \"\$kexec_script_tmp\" 2>&1 | tee /root/kexec/output.log || true
```

src/nixos-anywhere.sh
```diff
@@ -694,21 +813,50 @@ TMPDIR=/root/kexec setsid --wait ${maybeSudo} /root/kexec/kexec/run --kexec-extr
...
+    # Execute the local upload command and check for success
+    local uploadExitCode
+    (
+      set +x
+      runSsh sh -c "$(printf '%q' "$remoteCommands")"
+    )
+    uploadExitCode=$?
 
-    "${localUploadCommand[@]}" | runSsh sh -c "$(printf '%q' "$remoteCommands")"
+    handleKexecResult $uploadExitCode "Upload"
```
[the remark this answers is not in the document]
Qubasa: It's done like this on upstream. So for this PR I will stick to it

src/nixos-anywhere.sh
```diff
@@ -518,23 +573,35 @@ importFacts() {
...
+  for var in isOs isArch isInstaller isContainer isRoot hasIpv6Only hasTar hasCpio hasSudo hasDoas hasWget hasCurl hasSetsid remoteHomeDir remoteLogFile; do
     if [[ -z ${!var} ]]; then
       abort "Failed to retrieve fact $var from host"
     fi
   done
   set -u
 
+  # Clean up the remote log file on exit
+  # shellcheck disable=SC2064
+  trap "runSshNoTty rm -f \"$remoteLogFile\"" EXIT
```
Mic92, line 599: Does this not override our first trap?

src/nixos-anywhere.sh
```diff
@@ -694,21 +812,50 @@ TMPDIR=/root/kexec setsid --wait ${maybeSudo} /root/kexec/kexec/run --kexec-extr
...
+    # In the case where the ssh user is not root, we need to upload the kexec tarball
+    # to a location where the user has write permissions. We then use sudo to run
+    # kexec from that location.
+    if [[ -z $remoteHomeDir ]]; then
+      abort "Could not determine home directory for user $sshUser"
+    fi
+
+    (
+      set +x
+      "${localUploadCommand[@]}" | runSsh "cat > \"$remoteHomeDir\"/kexec-tarball.tar.gz"
```
Mic92, line 843: Why do we need a fact for this. Can we not always use ~/kexec-tarball.tar.gz?

Qubasa: @Mic92 No because the ssh user and the script user can be different.

src/get-facts.sh
```diff
@@ -20,4 +20,6 @@ hasWget=$(has wget)
 hasCurl=$(has curl)
 hasSetsid=$(has setsid)
 hasNixOSFacter=$(command -v nixos-facter >/dev/null && echo "y" || echo "n")
+remoteHomeDir=$HOME
+remoteLogFile=$(mktemp /tmp/nixos-anywhere-remote-log.XXXXXX)
```
Mic92, line 24: We can simplify this by also writing to `~/.nixos-anywhere.log` than we also get rid of the trap.

src/nixos-anywhere.sh
Qubasa: right yeah in this case you are right

src/nixos-anywhere.sh
```diff
@@ -58,16 +58,65 @@ hasWget=
 hasCurl=
 hasSetsid=
 hasNixOSFacter=
+remoteHomeDir=
+remoteLogFile=
 
 tempDir=$(mktemp -d)
-trap 'rm -rf "$tempDir"' EXIT
+trap 'rm -rf "$tempDir" && (runSshNoTty  -o ConnectTimeout=10 rm -f "$remoteLogFile" || true)' EXIT
 mkdir -p "$tempDir"
```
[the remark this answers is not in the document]
Qubasa: not true I just checked manually the remoteLogFile get's cleaned up correctly if I inject an exit 1 right after importFacts

provokateurin:
This PR broke custom installer images and using a user other than root: https://github.com/nix-community/nixos-anywhere/issues/614

PR #574, opened 2025-07-30
nixos-anywhere: Install jq when using --no-disko-deps

hgl:
This is a different way of fixing https://github.com/nix-community/disko/pull/1091

Nixos-anywhere will run commands like:

```
jq -r -f /nix/store/fpwn44vygjj6bfn8s1jj9p8yh6jhfxni-disk-deactivate/zfs-swap-deactivate.jq
...
jq -r --arg disk_to_clear /dev/sda -f /nix/store/fpwn44vygjj6bfn8s1jj9p8yh6jhfxni-disk-deactivate/disk-deactivate.jq
```
During installation, and with --no-disko-deps, `jq` isn't available on the official nixos-installer, and these command can fail.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -719,7 +719,11 @@ runDisko() {
   fi
 
   step Formatting hard drive with disko
-  runSsh "$diskoScript"
+  local installDeps=""
+  if [[ ${diskoDeps} == "n" ]]; then
+    installDeps="nix --extra-experimental-features 'nix-command flakes' profile install nixpkgs#jq &&"
+  fi
+  runSsh "$installDeps$diskoScript"
 }
 
 nixosInstall() {


hgl:
Thanks for the tip.

In my case, the remote machine is in a cloud environment that
1. Has very limited amount of ram
2. Can only mount predetermined official nixos iso

Which requires `--no-disko-deps`.

I send the PR because it seems the official installer already contains every tool needed by nixos-anywhere other than jq, and I assumed if jq can be added, we can use `--no-disko-deps` transparently at least with the official installer, but maybe this is isn't true?

I'm writing a downstream tool that uses nixos-anywhere to install nixos to remote machines, and I hoped to make the official installer a first class citizen where even if the ram is limited, nixos could still be successfully installed without user intervention. I wonder if this goal is feasible? 

If the design goal is that no deps disko should not trigger network download, which makes sense, then maybe it's not?

I've fixed the build error though.

Mic92:
Could you check if jq is included in the nixos installer closure? Maybe we can just expose jq?

Mic92:
Ah maybe extraDependencies doesn't get propageted into the $PATH. But in that case we might as well just include it environment.systemPackages as well.

Mic92:
Okay, than please open a pull request in nixpkgs and link it here.

hgl:
https://github.com/NixOS/nixpkgs/pull/429759

hgl:
PR has been merged.

PR #575, opened 2025-07-31
 nixos-anywhere: fix running `nixos-facter` when it isn't installed

Enzime:
This PR reverts #570 as I think we still want to run `nixos-facter` with `sudo` if possible and the changing the order of the flags doesn't fix the quoting issues

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -586,9 +586,14 @@ generateHardwareConfig() {
     fi
     if [[ ${hasNixOSFacter} == "n" ]]; then
       step "Generating facter.json using nixos-facter from nixpkgs"
-      runSshNoTty -o ConnectTimeout=10 ${maybeSudo} \
-        nix run "${nixOptions[@]}" \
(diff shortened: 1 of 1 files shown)


Enzime:
Tested this locally and it works 👍

src/nixos-anywhere.sh
```diff
@@ -586,11 +586,17 @@ generateHardwareConfig() {
     fi
     if [[ ${hasNixOSFacter} == "n" ]]; then
       step "Generating facter.json using nixos-facter from nixpkgs"
+
+      # We need to quote all the flags before they get passed to SSH
+      # otherwise SSH will drop the quotes which is necessary for
+      # `--extra-experimental-features "nix-command flakes"`.
```
sedlund, line 592: `--extra-experimental-features` can be passed multiple times.  related: https://github.com/NixOS/nix/issues/7789#issuecomment-2360234257

a-kenji:
Ah, I found the issue.

Enzime:
@a-kenji can you review this?

The latest commit should supercede #576

PR #576, opened 2025-07-31
nixos-anywhere: Fix installer check for nixos-facter invocation

a-kenji:
If we are not running in an installer we can't predict if
the system actually has access to sudo.
So we fall back to the non-sudo `nixos-facter` report.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -581,7 +581,7 @@ generateHardwareConfig() {
   mkdir -p "$(dirname "$hardwareConfigPath")"
   case "$hardwareConfigBackend" in
   nixos-facter)
-    if [[ ${isInstaller} == "y" ]]; then
+    if [[ ${isInstaller} == "n" ]]; then
       maybeSudo=""
     fi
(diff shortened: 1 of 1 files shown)


src/nixos-anywhere.sh
```diff
@@ -581,7 +581,7 @@ generateHardwareConfig() {
   mkdir -p "$(dirname "$hardwareConfigPath")"
   case "$hardwareConfigBackend" in
   nixos-facter)
-    if [[ ${isInstaller} == "y" ]]; then
+    if [[ ${isInstaller} == "n" ]]; then
```
Enzime, line 584: Instead of using the existing facts, I think we should gather the facts again after kexec'ing

Then we could check if we are in installer and determine if the kexec failed and throw an error

Enzime: I don’t think this PR is correct, if someone was not in an installer pre-kexec says nothing about whether `sudo` is available post-kexec

Enzime:
Should be fixed in #575

PR #577, opened 2025-08-01
Fully detach backgrounded reboot to prevent hanging SSH session

ciderale:
This PR ensures that the backgrounded reboot command is fully detached, preventing SSH from occasionally hanging after the command is sent. By redirecting both stdout and stderr, the ssh session can close cleanly without waiting on open channels. This project is really cool and I hope to contribute a small improvement.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -789,7 +789,7 @@ nixosReboot() {
     swapoff -a
     zpool export -a || true
   fi
-  nohup sh -c 'sleep 6 && reboot' >/dev/null &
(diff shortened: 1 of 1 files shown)


Enzime:
Out of curiosity, have you observed it hanging much and is it when something writes to stderr?

Enzime:
Interesting, I wonder if `sleep 6` is no longer necessary then @Mic92

Mic92:
> Interesting, I wonder if `sleep 6` is no longer necessary then @Mic92

The 6s were there so that we get a clean exit status and the kexec doesn't break the connection before the command finishes. I picked the rather high timeout in case we have a slow connection i.e. tor.

PR #600, opened 2025-11-03
feat: Add check and warning for reboot inhibitor

fidgetingbits:
In the event that there is some systemd inhibitor in place, the use of `2>&1`
from commit 4d55f9e (#577), hides the error: 'Call to Reboot failed: Access
denied'. To address that, I've added a check for inhibitors and a more explicit
warning.

In the event of an inhibitor being present, the new output is:

```
### Rebooting ###
Pseudo-terminal will not be allocated because stdin is not a terminal.
WARNING: Reboot is currently inhibited. Manual reboot may be required. Run 'systemd-inhibit --list' on the target details
Pseudo-terminal will not be allocated because stdin is not a terminal.
### Waiting for the machine to become unreachable due to reboot ###
```

If `--debug` is in use, then it will also print the full list of inhibitors
from `systemd-inhibit --list`, for example:

```
+ [[ 1 == 1 ]]
+ nixosReboot
+ step Rebooting
+ echo '### Rebooting ###'
### Rebooting ###
+ local inhibited
+ set +x
Pseudo-terminal will not be allocated because stdin is not a terminal.
WARNING: Reboot is currently inhibited. Manual reboot may be required. Run 'systemd-inhibit --list' on the target for further details
SSH COMMAND: ssh -t -o IdentitiesOnly=yes -i /tmp/tmp.vTDgb8CP5H/nixos-anywhere -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p 10022 root@moth.memeoid.cx sh

WHO            UID  USER  PID  COMM            WHAT                                                     WHY                                       MODE
Disk Manager   0    root  1659 udisksd         shutdown:sleep:idle                                      Unknown (mdraid-sync-job)                 block
ModemManager   0    root  1850 ModemManager    sleep                                                    ModemManager needs to reset devices       delay
NetworkManager 0    root  1119 NetworkManager  sleep                                                    NetworkManager needs to turn off networks delay
UPower         0    root  1532 upowerd         sleep                                                    Pause device polling                      delay
GNOME Shell    1001 nixos 1431 .gnome-shell-wr sleep                                                    GNOME needs to lock the screen            delay
GNOME Shell    1001 nixos 1431 .gnome-shell-wr sleep                                                    GNOME needs to save screen time data      delay
nixos          1001 nixos 1554 .gsd-media-keys handle-power-key:handle-suspend-key:handle-hibernate-key GNOME handling keypresses                 block
nixos          1001 nixos 1554 .gsd-media-keys sleep                                                    GNOME handling keypresses                 delay
nixos          1001 nixos 1556 .gsd-power-wrap sleep                                                    GNOME needs to lock the screen            delay

9 inhibitors listed.
+ runSsh sh
+ set +x
```

We could consider a new option that lets you to force `systemctl reboot -i` to
ignore inhibitors during reboot, but I haven't included that for now. Also I
don't bother exiting if an inhibitor is detected because we don't know for
certain atm the inhibitor runs long enough to actually stop the subsequent
reboot, so seems worth trying.

On example where there might be an inhibitor is if you are installing a raid
array using mdadm it can automatically trigger a multi-hour (dependent on disk
size) scan after initial configuration of the by disko.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -914,6 +914,29 @@ SSH
 
 nixosReboot() {
   step Rebooting
+
+  # disable debug output temporarily to prevent log spam
+  set +x
+  
+  local inhibited
+  inhibited=$(
+    runSsh sh <<SSH
+  if command -v systemd-inhibit 2>/dev/null; then 
+    systemd-inhibit --list | grep "shutdown"
+  fi
+SSH
+  )
+  if [[ -n "${inhibited}" ]]; then
+    echo "WARNING: Reboot is currently inhibited. Manual reboot may be required. Run 'systemd-inhibit --list' on the target for further details" 2>&1
+    if [[ -n ${enableDebug} ]]; then
+      echo "${inhibited}" 2>&1
+    fi
+  fi
+  
+  if [[ -n ${enableDebug} ]]; then
+    set -x
+  fi
+
   runSsh sh <<SSH
   if command -v zpool >/dev/null && [ "\$(zpool list)" != "no pools available" ]; then
     # we always want to export the zfs pools so people can boot from it without force import


src/get-facts.sh
```diff
@@ -19,5 +19,7 @@ hasDoas=$(has doas)
 hasWget=$(has wget)
 hasCurl=$(has curl)
 hasSetsid=$(if [ "$(has setsid)" = "y" ] && setsid --wait true 2>/dev/null; then echo "y"; else echo "n"; fi)
+hasGrep=$(has grep)
+hasSystemdInhibit=$(has systemd-inhibit)
```
Mic92: Both these checks are done in the non-kexeced system, but we than run the inhibit checked in a potential different system (i.e. the kexec image). So I don't think we need to do this check here.

PR #603, opened 2025-11-27
Add options to override default substituters and trusted-public-keys

leonardolang:
This commit introduces a way to override the default substituters and trusted-public-keys copied from the reference machine.

A common use case is when the reference machine points to a local "proxy" binary cache (e.g., 127.0.0.1), which will not work on the target host.

Using `--no-use-machine-substituters` disables the feature altogether, while `--option substituters <list>` and `--option extra-substituters <list>` results in errors due to the user not being part of trusted-users.

This commit also changes how `~/.config/nix/nix.conf` is created: it now stores a backup (with a `.orig` suffix) before patching the file, and reuses that backup on future invocations.

This avoids the situation where multiple runs of `nixos-anywhere` repeatedly append the same configuration over and over.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -22,6 +22,8 @@ nixOptions=(
 )
 SSH_PRIVATE_KEY=${SSH_PRIVATE_KEY-}
 machineSubstituters="y"
+substituters=""
+trustedPublicKeys=""
 
 declare -A phases
 phases[kexec]=1
@@ -167,6 +169,10 @@ Options:
   implies --no-use-machine-substituters
 * --no-use-machine-substituters
   don't copy the substituters from the machine to be installed into the installer environment
+* --substituters
+  override default substituters copied to the installer environment
+* --trusted-public-keys
+  override default trusted public keys copied to the installer environment
 * --debug
   enable debug output
 * --show-trace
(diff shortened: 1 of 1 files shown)


oneingan:
Thanks, we just hit a very similar real-world case, but with an SSH substituter rather than a localhost HTTP proxy.

Our reference machine config contains something like:

```nix
nix.settings.substituters = [
  \"https://nix-community.cachix.org\"
  \"ssh://eu.nixbuild.net\"
];
```

During install, `nixos-anywhere` copies those substituters into the temporary installer's `~/.config/nix/nix.conf`. In the installer we observed:

```ini
extra-substituters = ... ssh://eu.nixbuild.net ...
```

but the installer did not have the corresponding SSH context:
- no `known_hosts` / host CA for `eu.nixbuild.net`
- no `IdentityFile` / private key for that host
- no SSH config matching the final system

The result was a failure during the `disko` phase:

```text
Host key verification failed.
error: failed to start SSH connection to 'eu.nixbuild.net'
```

From inside the installer, a plain `ssh eu.nixbuild.net` failed with host key verification, and with host checks disabled it failed with `Permission denied (publickey)`, so this is not just a missing `known_hosts` entry; the installer lacks the full SSH trust/auth context.

Notably this happened even with `--build-on local`, because the temporary installer still ends up using its own substituter configuration while fetching/copying closures.

Because of that, I think the safe default should be to filter out SSH-based substituters when importing machine substituters into the installer, unless the user explicitly opts in. Copying `https://` caches by default makes sense, but copying `ssh://` caches/store URIs seems fragile because to make them work you also need to transfer:
- private key / agent state
- host key or host certificate CA
- matching SSH config / per-host options

`--no-use-machine-substituters` is currently too coarse, since it also disables perfectly valid HTTP(S) caches.

So +1 on making this configurable, and I would even argue for a default behavior like:
- copy `http://` / `https://` substituters
- drop `ssh://` (and possibly other SSH-backed store URIs) unless explicitly requested

That seems much safer for temporary installer environments.

PR #606, opened 2025-12-18
nixos-anywhere: Fix parsing of `--ssh-store-setting` without skipping next arg

a-kenji:
Fix parsing of the `--ssh-store-setting` without skipping the next
argument.

There was an extra shift in the `--ssh-store-setting`, also adjusted the
shifts to be in line with the others.

Credit to @name-snrl for the report, thank you.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -280,10 +280,10 @@ parseArgs() {
       ;;
     --ssh-store-setting)
       key=$2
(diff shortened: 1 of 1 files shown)


Enzime:
There's merge conflicts, can you rebase?

Enzime:
Already fixed in #607 but I do like that this PR still makes it consistent with the rest of the script

PR #623, opened 2026-02-13
Move kexec CLI checks inside the kexec branch

hgl:
I tried these changes inside a very minimal kexec env that I already booted into, where CLIs that are used to run kexec are not needed, and it can install without failing.


a-kenji (approved):
LGMT, did not test it yet.

PR #624, opened 2026-02-20
feat(nixos-install TF module): Add non sensite option

oliverwiegers:
Make it possible to print out nixos istall output despite of containing sensitive values.


inkules:
Good one! We have need for this, too!

PR #625, opened 2026-03-07
Fix SSH argument handling for timeouts and complex options

cafkafk:
This PR fixes two related issues with how SSH arguments are propagated, which were causing failures for deployments using complex SSH options (e.g., `ProxyCommand` with arguments) or relying on `ConnectTimeout`.

### Issues Fixed

**1. Fix `NIX_SSHOPTS` Quoting**
`sshArgs` was being flattened into `NIX_SSHOPTS` using `"${sshArgs[*]}"`. This naive expansion destroys the quoting of array elements that contain spaces.

Passing `--ssh-option "ProxyCommand='wrapper.sh' %h %p"` would result in `nix` splitting the proxy command string, causing SSH to execute the wrapper without arguments (failed IAP tunnels, etc.).

**Fix:** Changed to `NIX_SSHOPTS="$(printf '%q ' "${sshArgs[@]}")"` to properly shell-escape the arguments, preserving the structure of complex flags when passed to `nix`.

**2. Fix `ConnectTimeout` Injection**
Internal calls like `runSsh -o ConnectTimeout=10` were broken because `runSsh` appends positional arguments (`$@`) *after* the destination.

`ssh root@host -o ConnectTimeout=10` tries to execute `-o` as a **remote command** on the server. This returns exit code 127 (command not found) instead of actually setting a connection timeout, breaking logic that relies on exit codes (like the post-kexec wait loop).

**Fix:** Refactored these calls to append `-o ConnectTimeout=10` to the `sshArgs` array (scoped within subshells where appropriate) so they are correctly passed as SSH flags *before* the destination.

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -481,13 +481,13 @@ runSsh() {
 }
 
 nixCopy() {
-  NIX_SSHOPTS="${sshArgs[*]}" nix copy \
+  NIX_SSHOPTS="$(printf '%q ' "${sshArgs[@]}")" nix copy \
     "${nixOptions[@]}" \
     "${nixCopyOptions[@]}" \
     "$@"
 }
 nixBuild() {
-  NIX_SSHOPTS="${sshArgs[*]}" nix build \
+  NIX_SSHOPTS="$(printf '%q ' "${sshArgs[@]}")" nix build \
     --print-out-paths \
     --no-link \
     "${nixBuildFlags[@]}" \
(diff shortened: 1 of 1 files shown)


cafkafk:
Shellcheck [here](https://github.com/nix-community/nixos-anywhere/compare/d2821205271711095f29d72cb314760df69e701e..ac10e25d02f54228faf07eb56621c1fd16b45dc9) and [here](https://github.com/nix-community/nixos-anywhere/compare/ac10e25d02f54228faf07eb56621c1fd16b45dc9..bbb2cc2af9277949b4bcefba4b444092d06f5e98) were false positives, so I've silence them.

cafkafk:
For context: The failing tests seems to be an issue with CI, not this PR.

PR #627, opened 2026-03-24
--post-kexec-ssh-host as argument (equivalent to --post-kexec-ssh-port)

findus:
This fixes an issue i have when running nixos-anywhere on devices in an environment that doesn't have a dhcp server and static IPs configured.

With that command and a custom kexec tarball that enables link lokal ipv6 addresses I can connect to the link local ipv6 address after kexec.

Example:

```
nix run github:findus/nixos-anywhere -- --post-kexec-ssh-host 'fe80::ffff:ffff:ffff:93b0%eth0' --kexec "result/nixos-kexec-installer-x86_64-linux.tar.gz" --flake '.#myflake' root@192.168.69.194
```

### DIFF
--- a/src/nixos-anywhere.sh
+++ b/src/nixos-anywhere.sh
@@ -39,6 +39,7 @@ else
 fi
 sshConnection=
 postKexecSshPort=22
+postKexecSshHost=
 buildOnRemote=n
 buildOn=auto
 envPassword=n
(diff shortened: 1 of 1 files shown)


phaer:
Just curious & trying to understand the use-case in more detail: You need to have root on the pre-kexec system anyway (to kexec), so you should be able to use EUI-64 link-local ipv6 addresses for both, the pre- and the post-kexec connection, no?

findus:
It made my life a little bit easier by not having to manually enable link local ipv6 addresses in NetworkManager/networkd/etc. on installations I want to do kexec on.

But yes, I could also do that and tunnel-visioned a little too hard on this solution I guess.

PR #636, opened 2026-05-15
nixos-anywhere: add flag to kexec in the installer

jackrosenberg:
Adds a flag to force kexec in the installer, for when kernel/userspace is running from the same device as the one we want to install to. Thanks to @Lassulus for the suggestion.


Lassulus (approved):
lgtm

PR #642, opened 2026-05-26
nixos-anywhere: bump default kexec image to nixos-25.11

Mic92:
nixos-images released the 25.11 branch. Bump the default kexec installer URL so new installs pick up the current channel by default.


Mic92:
A bit late and I soon have to do it again...

